PR #16, opened 2017-11-10
SSCS-2244 Convert SYA Json to Case

jackmaloney:
- Created a Deserializer to convert SYA Json to a CCD Case object
- Front end still in development so made assumptions on some field names and some are missing. These are marked with TODOs
- Added Equals and Hashcode methods for the POJOs


aaronrs:
Pairing Review

PR #29, opened 2018-01-09
Sscs 2526 sya pdf generation

sathishbabuj:


src/main/resources/templates/appellant_appeal_template.html
```diff
@@ -0,0 +1,542 @@
...
+
+                <div>
+                    <dt class="cya-question">Dates you're not available</dt>
+                    <dd class="cya-answer">
+                        {% if SyaCaseWrapper.hearing.scheduleHearing == true  %}
+                        {% for date in SyaCaseWrapper.hearing.datesCantAttend %}
+                        {{ date }}<br/>
+                        {% endfor %}
+                        {% else %}
+
```
jackmaloney: Can you make sure the latest changes to the TYA API appellant_appeal_template.html are copied over?

src/main/resources/config/application.properties
```diff
@@ -19,7 +19,7 @@ appeal.email.port=${SPRING_MAIL_PORT:25}
 appeal.email.smtp.tls.enabled=${SPRING_MAIL_PROPERTIES_MAIL_SMTP_STARTTLS_ENABLE:true}
 appeal.email.smtp.ssl.trust=${SPRING_MAIL_SMTP_SSL_TRUST:*}
 
-idam.s2s-auth.totp_secret=${IDAM_S2S_AUTH_TOTP_SECRET:J7PBT37Y6C3SEFGX}
+idam.s2s-auth.totp_secret=${IDAM_S2S_AUTH_TOTP_SECRET:secret}
```
jackmaloney: Is it safe to have this as part of the commit history? Might be worth rebasing if not

src/main/resources/templates/appellant_appeal_template.html
sathishbabuj: Not now later. I copied the latest template but the PDF generation was failing because of some exception in the date format.

src/main/resources/config/application.properties
sathishbabuj: This branch will be deleted.

PR #40, opened 2018-01-16
Feature/sscs 2524 retrieve case from ccd

savirneni:
SSCS-2524:[SA]: Retrieve Case from CCD given a TYA Appeal Number


src/main/java/uk/gov/hmcts/sscs/domain/corecase/CcdCaseResponse.java
```diff
@@ -0,0 +1,156 @@
...
+    private AfterSubmitCallbackResponse afterSubmitCallbackResponse;
+
+    @JsonProperty(value = "callback_response_status")
+    private String callbackResponseStatus;
+    
+    @JsonProperty(value = "callback_response_status_code")
+    private String callbackResponseStatusCode;
+
+    @JsonProperty(value = "case_data")
+    private CcdCase caseData;
```
jackmaloney: Is this just a placeholder? I don't think this is going to work as the names in the `case_data` json do not match with CcdCase. This would involve creating a new deserialiser which converts the Json into the relevant object.

savirneni: yes it is just a place holder. We need to match the names in case_data json with the CcdCase.

PR #80, opened 2018-02-19
Env vars

bartlabno:
Infrastructure update, vars per env, remove sensitive data and moved into vault

### DIFF
--- a/infrastructure/main.tf
+++ b/infrastructure/main.tf
@@ -1,10 +1,4 @@
 provider "vault" {
-  //  # It is strongly recommended to configure this provider through the
-  //  # environment variables described above, so that each user can have
-  //  # separate credentials set in the environment.
-  //  #
-  //  # This will default to using $VAULT_ADDR
-  //  # But can be set explicitly
   address = "https://vault.reform.hmcts.net:6200"
 }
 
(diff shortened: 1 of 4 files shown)


infrastructure/main.tf
```diff
@@ -1,27 +1,61 @@
...
 
 data "vault_generic_secret" "email_mac_secret" {
-  path = "secret/test/sscs/sscs_email_mac_secret_text"
+  path = "secret/${var.infrastructure_env}/sscs/sscs_email_mac_secret_text"
 }
 
-data "vault_generic_secret" "sscs_rpc_email_id" {
-  path = "secret/test/sscs/secret/test/sscs/sscs_rpc_email_id"
+data "vault_generic_secret" "sscs_tribunals_case_secret" {
+  path = "secret/${var.infrastructure_env}/ccidam/service-auth-provider/api/microservice-keys/sscs-tribunals-case"
```
sathishbabuj: There can be more than one secret in Tribunal case api. It will be good to make name specific instead of sscs_tribunals_case_secret

bartlabno: This value was introduced on previous commits and it links to idam vaults not sscs

bartlabno: Replaces https://github.com/hmcts/tribunals-case-api/pull/80/files#diff-1031eeab417cd8afe77e7e005717d38dL24
which allows to use different envs only

sathishbabuj: This is for the vault key name **data "vault_generic_secret" "sscs_tribunals_case_secret"** and not the value.

PR #86, opened 2018-02-26
SSCS-2690[SA]:Implement logging in app insights on tribunals api

savirneni:
Implement logging in app insights on tribunals api

https://tools.hmcts.net/jira/browse/SSCS-2690

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -104,6 +104,7 @@ dependencies {
 
     compile group: 'uk.gov.hmcts.reform', name: 'java-logging-spring', version: '1.4.2'
     compile group: 'uk.gov.hmcts.reform', name: 'java-logging-httpcomponents', version: '1.4.2'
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -104,6 +104,7 @@ dependencies {
 
     compile group: 'uk.gov.hmcts.reform', name: 'java-logging-spring', version: '1.4.2'
```
iguannago: move to 1.6.1 as well.

build.gradle
```diff
@@ -104,6 +104,7 @@ dependencies {
...
     compile group: 'uk.gov.hmcts.reform', name: 'java-logging-httpcomponents', version: '1.4.2'
```
iguannago: move to 1.6.1

build.gradle
savirneni: Updated to correct version.

build.gradle
savirneni: Updated to correct version.

PR #90, opened 2018-02-26
Update PDF for hearing arrangements section

bendiggle:
- Add new fields `interpreterLanguageType`, `signLanguageType`
- Add other `option`
- Change `disabledAccess` to `accessibleHearingRoom`

**Example:**
- If `sign language interpreter` has been selected and the `signLanguageType` field has a value then display that value else display `Required`. Display `Not required` when sign language interpreter hasn't been selected.

### DIFF
--- a/src/main/resources/templates/appellant_appeal_template.html
+++ b/src/main/resources/templates/appellant_appeal_template.html
@@ -314,10 +314,27 @@
         </div>
 
         <div>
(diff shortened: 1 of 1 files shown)


bendiggle:
This is ready to be reviewed now

screen shot 2018-03-27 at 13 15 33

PR #141, opened 2018-04-13
SSCS-2926 Send reminder event to CCD

jackmaloney:
- Create endpoint for reminders
- Send reminder to CCD as an event

### DIFF
--- a/src/main/java/uk/gov/hmcts/sscs/service/ccd/ReadCoreCaseDataService.java
+++ b/src/main/java/uk/gov/hmcts/sscs/service/ccd/ReadCoreCaseDataService.java
@@ -26,7 +26,7 @@ public class ReadCoreCaseDataService {
     }
 
     public CaseDetails getCcdCaseDetailsByCaseId(String caseId) {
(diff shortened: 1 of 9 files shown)


src/main/java/uk/gov/hmcts/sscs/service/ccd/ReadCoreCaseDataService.java
```diff
@@ -26,7 +26,7 @@ public ReadCoreCaseDataService(CoreCaseDataService coreCaseDataService) {
     }
 
     public CaseDetails getCcdCaseDetailsByCaseId(String caseId) {
-        EventRequestData eventRequestData = coreCaseDataService.getEventRequestData("appealCreated");
+        EventRequestData eventRequestData = coreCaseDataService.getEventRequestData(null);
         String serviceAuthorization = coreCaseDataService.generateServiceAuthorization();
```
sathishbabuj: Why null parameter in the request?

jackmaloney: `appealCreated` is never used. It should have been removed a while ago so I was just cleaning this up.

PR #144, opened 2018-04-30
SSCS-3105 Fixed the code to handle mulitple hearing booked events whi…

sathishbabuj:
…ch has only latest hearing booked details


Edlyne:
Verified fixed and multiple other scenarios tested. Confirmed all PASSED and the Jenkins build/tests all passed.

PR #149, opened 2018-05-21
SSCS-2877 User id extracted from IDAM authorization token

fantasticjamieburns:

https://tools.hmcts.net/jira/browse/SSCS-2877

### Change description ###

IDAM user_id is taken from the Authorisation token, instead of being loaded from application configuration.

```
[ ] Yes
[x] No
```


fantasticjamieburns:
Yes, @iguannago a shared jar might be useful given we do the same thing in 3 places.
I think another story though would be good so it can be prioritized against other work?

PR #151, opened 2018-05-30
Feature/sscs 3255 send email when ccd down

iguannago:
This change covers this ticket https://tools.hmcts.net/jira/browse/SSCS-3255

Basically, it's about sending out the PDF by email even if CCD is unavailable. 

**Please, take into account you only have to review the last two commits.**


src/main/java/uk/gov/hmcts/sscs/service/SubmitAppealService.java
```diff
@@ -59,38 +55,51 @@
     }
 
     public void submitAppeal(SyaCaseWrapper appeal) {
-        String appellantLastName = appeal.getAppellant().getLastName();
-        String nino = appeal.getAppellant().getNino();
-
-        String appellantUniqueId = String.format(ID_FORMAT, appellantLastName, nino.substring(nino.length() - 3));
+        CaseData caseData = transformAppealToCaseData(appeal);
+        CaseDetails caseDetails = createCaseInCcd(caseData);
+        byte[] pdf = generatePdf(appeal, caseDetails.getId());
```
sathishbabuj: A null check needs to be done for caseDetails.getId() if there is any error while storing case to CCD.

src/main/java/uk/gov/hmcts/sscs/service/SubmitAppealService.java
```diff
@@ -1,72 +1,102 @@
...
+        sendPdfByEmail(appeal, pdf);
+    }
+
+    private CaseDetails createCaseInCcd(CaseData caseData) {
+        try {
+            return ccdService.createCase(caseData);
+        } catch (CcdException ccdEx) {
```
savirneni: AppInsight alert missing when CCD Exception is thrown.

src/main/java/uk/gov/hmcts/sscs/service/SubmitAppealService.java
```diff
@@ -1,72 +1,102 @@
...
+    }
+
+    private void sendPdfByEmail(SyaCaseWrapper appeal, byte[] pdf) {
         String appellantLastName = appeal.getAppellant().getLastName();
         String nino = appeal.getAppellant().getNino();
-
         String appellantUniqueId = String.format(ID_FORMAT, appellantLastName, nino.substring(nino.length() - 3));
+        submitYourAppealEmail.setSubject(appellantUniqueId);
+        submitYourAppealEmail.setAttachments(newArrayList(pdf(pdf, appellantUniqueId + ".pdf")));
+        emailService.sendEmail(submitYourAppealEmail);
```
savirneni: AppInsight alert missing when emailService is not available.

src/main/java/uk/gov/hmcts/sscs/service/SubmitAppealService.java
iguannago: @savirneni , I just configured it as it is for the caseloader which is cleaner btw. And the caseloader is receiving appInsight alerts successfully. The way the code was configured before is not consided best practice. More details here -> https://stackify.com/best-practices-exceptions-java/

src/main/java/uk/gov/hmcts/sscs/service/SubmitAppealService.java
iguannago: It's not missing. There is an uchecked exception that will be triggered if there is any issue when sending an email. That will generate an AppInsight alert. Don't see any issue here. Same as above comment, really.

src/main/java/uk/gov/hmcts/sscs/service/SubmitAppealService.java
savirneni: If we receive alert then thats fine.

src/main/java/uk/gov/hmcts/sscs/service/SubmitAppealService.java
savirneni: If we receive alert then thats fine.

PR #152, opened 2018-05-30
Task/sscs 3227 tribunals api docker

savirneni:
New dockerfile added for the DockerHub.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ Once the application running locally, please make sure
     For example:  export TEST_URL=http://localhost:8080
 3. Execute ./gradlew --info smoke
 
-### Running in Docker
+### Running in Docker(Work in progress...)
 Create the image of the application by executing the following command:
 
 ```


Dockerfile
```diff
@@ -1,11 +1,11 @@
...
-COPY build/install/tribunals-case-api /opt/app/
+COPY build/libs/tribunals-case-api.jar /opt/app/
 
 WORKDIR /opt/app
 
-HEALTHCHECK --interval=10s --timeout=10s --retries=10 CMD http_proxy="" curl --silent --fail http://localhost:8083/health
+HEALTHCHECK NONE
 
-EXPOSE 8083
+EXPOSE 8082
```
sathishbabuj: sscs case loader uses port 8082. So it will be better to use TCA API default port of 8080

hub.Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+
+FROM openjdk:8-jre-alpine
+
+COPY --from=builder /home/gradle/src/build/libs/tribunals-case-api.jar /opt/app/
+
+WORKDIR /opt/app
+
+HEALTHCHECK NONE
+
+EXPOSE 8082
```
sathishbabuj: sscs case loader uses port 8082. So it will be better to use TCA API default port of 8080

Dockerfile
savirneni: Port updated to use 8080

hub.Dockerfile
savirneni: Port updated to use 8080

PR #153, opened 2018-06-07
Sscs 3265 evidence document upload

sathishbabuj:


fantasticjamieburns:
I think we should be rebasing -- not merging changes -- from master.
And it looks like the merge has resulted in duplication of IDAM stuff.

Class `CoreCaseDataService` should not have `IdamService` reimplemented as protected methods...

PR #160, opened 2018-06-19
Sscs 3050 - Email robotics

jackmaloney:
- Generate Robotics Json based on SYA data
- Validate against the Robotics schema
- Email to the Robotics team

This work is behind a feature flag which can be toggled per environment, defaulted to false


src/main/java/uk/gov/hmcts/sscs/email/RoboticsEmail.java
```diff
@@ -0,0 +1,18 @@
...
+import org.springframework.stereotype.Component;
+
+@Component
+public class RoboticsEmail extends Email {
+
+    public RoboticsEmail(@Value("${appeal.email.from}") String from,
+                         @Value("${appeal.email.to}") String to,
+                         @Value("${appeal.email.subject}") String subject,
+                         @Value("${appeal.email.message}") String message) {
+        this.from = from;
```
sathishbabuj: The application properties should 
robotics.email.to
robotics.email.subject
robotics.email.message
robotics.email.enabled

src/main/java/uk/gov/hmcts/sscs/service/RoboticsService.java
```diff
@@ -0,0 +1,32 @@
...
+
+@Service
+public class RoboticsService {
+
+    private final InputStream inputStream;
+
+    private final Schema schema;
+
+    public RoboticsService() {
+        inputStream = getClass().getResourceAsStream("/schema/sscs-robotics.json");
```
fantasticjamieburns: This isn't a big thing I guess, but I don't think constructors should be used for anything other than accepting required dependencies and setting them into properties.
https://www.google.co.uk/search?q=constructors+should+not+do+work

Also, I would prefer the json generator to be a dependency injected into this coordinating class -- static methods bind calling code to a concrete implementation in an inflexible way and are not well suited to a growing project.

Finally, I would prefer the schema validator to be extracted out and dependency injected into this coordinating class too.

src/main/java/uk/gov/hmcts/sscs/json/RoboticsJsonGenerator.java
```diff
@@ -0,0 +1,139 @@
...
+
+        return obj;
+    }
+
+    private static JSONObject buildAppealDetails(JSONObject obj, SyaCaseWrapper appeal) {
+        obj.put("caseCode", "002DD");
+        obj.put("appellantNino", appeal.getAppellant().getNino());
+        //FIXME: To be implemented at a future date as part of another ticket
+        obj.put("appellantPostCode", "Bedford");
+        obj.put("appealDate", LocalDate.now().toString());
```
fantasticjamieburns: `LocalDate.now()` is a temp value until a future story?
If so, can we set it to something obviously incorrect, so we don't leave this in by mistake.

jackmaloney: No this isn't a temp value, this is correct

infrastructure/main.tf
```diff
@@ -55,6 +55,14 @@ data "vault_generic_secret" "appeal_email_to" {
...
 }
 
+data "vault_generic_secret" "robotics_email_from" {
+  path = "secret/${var.infrastructure_env}/sscs/robotics_email_from"
+}
+
+data "vault_generic_secret" "robotics_email_to" {
+  path = "secret/${var.infrastructure_env}/sscs/robotics_email_to"
+}
+
```
sathishbabuj: If we are going to use the same email from and to as Email pdf, then above two declaration is not needed.

src/main/java/uk/gov/hmcts/sscs/email/Email.java
```diff
@@ -2,7 +2,7 @@
 
 import java.util.List;
 
-public class Email {
+public abstract class Email {
```
sathishbabuj: Why abstract class?. Don't see any abstract method.

infrastructure/main.tf
fantasticjamieburns: i disagree about the vault value -- 
even if they are the same now, independent keys make the featured decoupled and clear about what they are being used for

src/main/java/uk/gov/hmcts/sscs/email/Email.java
fantasticjamieburns: Inheritance is anyway a bad design pattern, and this class cannot be used without a sub-class so the class is marked abstract to make that clear.

PR #167, opened 2018-07-03
SSCS-3333 Set RPC to Birmingham for no SC number

sathishbabuj:


src/main/java/uk/gov/hmcts/sscs/service/referencedata/RegionalProcessingCenterService.java
```diff
@@ -65,6 +65,11 @@ private void populateRpcMetadata() {
 
 
     public RegionalProcessingCenter getByScReferenceCode(String referenceNumber) {
+
+        if (null == referenceNumber) {
```
fantasticjamieburns: Could it be an empty string?

PR #174, opened 2018-07-11
Sscs 3421 transfrom pip office number

chrisgrimble:
SSCS-3421 Pass the PIP office number from the appeal to robotics if it 
has been set


chrisgrimble:
Have created a new branch and PR for this

PR #176, opened 2018-07-12
SSCS-3387 Compare surnames

chrisgrimble:
When we check the surname given by a user to track their
appeal they may use slightly different letters i.e. include/disclude
accents, slightly different hypens (often the case if copied from a word
document compared to being typed in) etc.

When we compare the surnames convert the charcters to ASCII and remove
anything that is not a lower or upper case letter before comparison.


src/main/java/uk/gov/hmcts/sscs/service/CcdService.java
```diff
@@ -122,10 +123,16 @@ public CaseData findCcdCaseByAppealNumberAndSurname(String appealNumber, String
...
                 && caseData.getAppeal().getAppellant().getName().getLastName() != null
-                && caseData.getAppeal().getAppellant().getName().getLastName().equalsIgnoreCase(surname)
+                && compareSurnames(surname, caseData)
                 ? caseData : null;
     }
 
+    private boolean compareSurnames(String surname, CaseData caseData) {
+        String caseDataSurname = caseData.getAppeal().getAppellant().getName().getLastName().replaceAll("[^a-zA-Z]", "");
+        String unidecodeSurname = unidecode(surname).replaceAll("[^a-zA-Z]", "");
+        return unidecode(caseDataSurname).equalsIgnoreCase(unidecodeSurname);
```
fantasticjamieburns: Shouldn't we also remove non alpha characters from the case data surname, to get a match?

PR #192, opened 2018-07-23
[PREVIEW] SSCS-3170 Store the PDF we send to gaps 2 in CCD so we can retrieve it

chrisgrimble:
later if it is needed.

Add the pdf we send to Gaps2 to the document store. Then update CCD
with the location so the pdf can be retrieved later if needed.
Don't fail if either adding the pdf to the document store or
updating CCD fails.


src/main/java/uk/gov/hmcts/sscs/service/PdfStoreService.java
```diff
@@ -0,0 +1,51 @@
...
+                    .documentFileName(fileName)
+                    .documentDateAdded(LocalDateTime.now().format(DateTimeFormatter.ISO_DATE))
+                    .documentLink(documentLink)
+                    .build();
+            SscsDocument pdfDocument = SscsDocument.builder().value(sscsDocumentDetails).build();
+
+            return Collections.singletonList(pdfDocument);
+        } catch (RestClientException exc) {
+            log.info("Failed to store pdf document but carrying on [" + fileName + "]");
+            log.debug("Exception from document store", exc);
```
sathishbabuj: It will be useful if the exception is also logged at info level. Do we need two log level?

src/main/java/uk/gov/hmcts/sscs/service/SubmitAppealService.java
```diff
@@ -87,6 +98,19 @@ private CaseDetails createCaseInCcd(CaseData caseData) {
         try {
             return ccdService.createCase(caseData);
         } catch (CcdException ccdEx) {
+            log.info("Failed to create ccd case but carrying on [" + caseData.getCaseReference() + "]");
+            log.debug("Exception from CCD", ccdEx);
```
sathishbabuj: It will be useful if the exception is also logged at info level. Do we need two log level?

akwarteng:
This has been QA verified

PR #196, opened 2018-07-26
SSCS-3662 Sending email is not thread safe

chrisgrimble:
The submit you appeal or robotics emails were not being constructed in a
thread safe manor. This could have lead to the emails contain details
from multiple appeals if multiple people are submiting their appeal at
once. Made it so we do not store the email data on the service.


src/main/java/uk/gov/hmcts/sscs/email/EmailAttachment.java
```diff
@@ -55,4 +56,24 @@ public String toString() {
         return String.format("EmailAttachment{contentType='%s', filename='%s'}",
                 filename, contentType);
     }
+
+    @Override
+    public boolean equals(Object o) {
```
iguannago: cannot you use lombok as well here -> https://projectlombok.org/features/EqualsAndHashCode

akwarteng:
This has been QA verified

PR #203, opened 2018-08-01
SSCS-3699 Change robotics extension from json to txt

jackmaloney:


akwarteng:
This has been QA verified

PR #222, opened 2018-09-19
[PREVIEW] Sscs 3932 evidence indicator robotics

jackmaloney:


src/main/java/uk/gov/hmcts/sscs/service/SubmitAppealService.java
```diff
@@ -98,7 +98,7 @@ public void submitAppeal(SyaCaseWrapper appeal) {
         prepareCaseForPdf(appeal, caseDetails.getId(), caseData, pdf, idamTokens);
 
         if (roboticsEnabled) {
-            sendCaseToRobotics(appeal, caseDetails.getId(), postcode, pdf);
+            sendCaseToRobotics(appeal, caseDetails.getId(), postcode, pdf, caseData.getEvidencePresent());
```
iguannago: this method is getting too many args. It's nicer to have fewer args. Maybe a simple class to store those props would be cleaner. No need to do this now, Jack, just saying it :)

PR #227, opened 2018-09-26
[PREVIEW] SSCS-3781 Removed vault_config.tf and updated Jenkinsfile_CNP file

sathishbabuj:


Jenkinsfile_CNP
```diff
@@ -11,19 +11,19 @@ def product = "sscs"
 def component = "tribunals-api"
 
 List<LinkedHashMap<String, Object>> secrets = [
-  secret('idam-url', 'IDAM_URL'),
+  secret('idam-api', 'IDAM_URL'),
   secret('s2s-api', 'IDAM.S2S-AUTH'),
```
iguannago: rename it to idam-s2s-api, please. this is the one used in caseloader so that we use the same secrets.

PR #263, opened 2018-10-12
[PREVIEW]Sscs 4122 tribunals api nulls

PaulVM:
Fix to prevent CCD being touched a second time if the first call fails to create a case.


iguannago:
@PaulVM , in the createCaseInCcd and updateCaseInCcd could you, please, change the exception being caught to Throwable so that we ensure that any exception is caught and we fall back to the sendEmailByPdf successfully?

PR #281, opened 2018-10-23
[PREVIEW]Fix test when two dates bridge daylight savings

chris-moreton:
When the start date is in a different daylight savings period to the calculated future date, the time parts of the two dates, expressed in UTC, should be different. Previously the test always expected the UTC parts to be the same.


iguannago:
I just prefixed it with the preview to run a complete batch of tests.

PR #331, opened 2018-11-23
SSCS-4283 write appointee details to ccd

arcotc:


src/test/resources/sya/allDetailsWithAppointeeWithSameAddress.json
```diff
@@ -0,0 +1,98 @@
...
+  "appellant": {
+    "title": "Mr",
+    "firstName": "Joe",
+    "lastName": "Bloggs",
+    "dob": "26-02-1990",
+    "nino": "AB877533C",
+    "isAddressSameAsAppointee": true,
+    "contactDetails": {
+      "phoneNumber": "07411222222",
+      "emailAddress": "sscstest@greencroftconsulting.com"
```
ryandac: I presume these contact details are always the same as the appointee for this scenario. Is that correct?

Why would the appellant enter different contact details if they didn't want notifications?

PR #343, opened 2018-12-10
SSCS-4614 - appointee tya number generation

ryandac:
Added appointee Subscription to subscriptions
Also contains: PR for  SSCS-4614 - TYA Appeal number generation for appointees

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/sscs/service/SubmitAppealService.java
+++ b/src/main/java/uk/gov/hmcts/reform/sscs/service/SubmitAppealService.java
@@ -1,5 +1,7 @@
 package uk.gov.hmcts.reform.sscs.service;
 
+import static uk.gov.hmcts.reform.sscs.transform.deserialize.SubmitYourAppealToCcdCaseDataDeserializer.convertSyaToCcdCaseData;
+
 import java.net.URI;
 import java.util.Collections;
 import java.util.LinkedHashMap;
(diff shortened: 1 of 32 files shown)


src/main/java/uk/gov/hmcts/reform/sscs/service/SubmitAppealService.java
```diff
@@ -111,42 +109,12 @@ private SscsCaseDetails createCaseInCcd(SscsCaseData caseData, IdamTokens idamTo
...
-        return updateCaseData(caseData);
-    }
-
-    protected SscsCaseData transformAppealToCaseData(SyaCaseWrapper appeal, String region, RegionalProcessingCenter rpc) {
-
-        SscsCaseData caseData = submitYourAppealToCcdCaseDataDeserializer.convertSyaToCcdCaseData(appeal, region, rpc);
-
-        return updateCaseData(caseData);
+    private SscsCaseData transformAppealToCaseData(SyaCaseWrapper appeal) {
+        return submitYourAppealToCcdCaseDataDeserializer.convertSyaToCcdCaseData(appeal, appealNumberGenerator);
```
iguannago: why do you need to pass the appealNumberGenerator as a parameter? it's not used anywhere else in this class, right? 
I would actually make the AppealNumberGenerator.generateAppealNumber() method static.

src/main/java/uk/gov/hmcts/reform/sscs/service/SubmitAppealService.java
```diff
@@ -111,42 +106,12 @@ private SscsCaseDetails createCaseInCcd(SscsCaseData caseData, IdamTokens idamTo
...
-        return updateCaseData(caseData);
-    }
-
-    protected SscsCaseData transformAppealToCaseData(SyaCaseWrapper appeal, String region, RegionalProcessingCenter rpc) {
-
-        SscsCaseData caseData = submitYourAppealToCcdCaseDataDeserializer.convertSyaToCcdCaseData(appeal, region, rpc);
-
-        return updateCaseData(caseData);
+    private SscsCaseData transformAppealToCaseData(SyaCaseWrapper appeal) {
+        return submitYourAppealToCcdCaseDataDeserializer.convertSyaToCcdCaseData(appeal);
```
iguannago: Could you make this method static? It does not hold any status and it does not need of any injection services, so in theory, it could be static and that will fix the sonar issue since we don't need to injected etc. Thanks

PR #344, opened 2018-12-10
SSCS-4702 Do not run certain case creation steps if duplicate is found

PaulVM:


PaulVM:
Not needed as per David Crespo's comment on SSCS-4702

PR #356, opened 2019-01-02
SSCS-4737 Make error message less misleading

PaulVM:
It doesn't carry on.


src/main/java/uk/gov/hmcts/reform/sscs/service/SubmitAppealService.java
```diff
@@ -98,7 +98,7 @@ private SscsCaseDetails createCaseInCcd(SscsCaseData caseData, IdamTokens idamTo
             }
         } catch (Exception e) {
             log.error("Error found in the case creation or callback process for ccd case with "
-                            + "Nino - {} and Benefit type - {} but carrying on ",
+                            + "Nino - {} and Benefit type - {} ",
```
srochani: There are two parameters in log.error but we are passing three. Please add another {}

PR #364, opened 2019-01-14
Store Robotics JSON with CCD case

chris-moreton:
Details of this change can be found at https://tools.hmcts.net/jira/browse/SSCS-4595


src/main/java/uk/gov/hmcts/reform/sscs/service/RoboticsJsonUploadService.java
```diff
@@ -0,0 +1,129 @@
...
+        log.info("Uploading Robotics JSON for case {}", caseDetails.getId());
+        UploadResponse uploadResponse = uploadRoboticsJson(singletonList(file));
+
+        log.info("Attaching Robotics JSON to case {}", caseDetails.getId());
+        SscsCaseData updatedCaseData = attachRoboticsJsonToCaseData(caseData, uploadResponse);
+
+        try {
+            ccdService.updateCase(updatedCaseData, caseDetails.getId(),
+                    "attachRoboticsJson", "", "", idamTokens);
+        } catch (CcdException ccdEx) {
```
iguannago: here we have to catch any Exception rather than only the CcdException and then carry on as you currently do. And I wouldn't log it as an ERROR since that will trigger an alert. I would log it as a log.info and then Thor can set an alert if he wishes. Please, check with him on that.

src/main/java/uk/gov/hmcts/reform/sscs/service/SubmitAppealService.java
```diff
@@ -67,6 +72,24 @@ private void postCreateCaseInCcdProcess(SyaCaseWrapper appeal, String firstHalfO
...
             roboticsService.sendCaseToRobotics(caseData, caseDetails.getId(), firstHalfOfPostcode, pdf,
                     additionalEvidence);
+
+            attachRoboticsJsonToCaseInCcd(caseData, idamTokens, caseDetails);
+        }
+    }
+
+    private void attachRoboticsJsonToCaseInCcd( SscsCaseData caseData,
+                                                IdamTokens idamTokens, SscsCaseDetails caseDetails) {
+        byte[] pdf = sscsPdfService.generateAndSendPdf(caseData, caseDetails.getId(), idamTokens);
```
iguannago: Why are we generating and sending the PDF again?

src/main/java/uk/gov/hmcts/reform/sscs/service/RoboticsJsonUploadService.java
```diff
@@ -0,0 +1,129 @@
...
+    public UploadResponse uploadRoboticsJson(List<MultipartFile> files) {
+
+        String serviceAuthorization = authTokenGenerator.generate();
+
+        try {
+            return documentUploadClientApi
+                    .upload(S2S_TOKEN, serviceAuthorization, DM_STORE_USER_ID, files);
+        } catch (HttpClientErrorException httpClientErrorException) {
+            log.error("Doc Store service failed to upload Robotics JSON", httpClientErrorException);
+            throw new UnsupportedDocumentTypeException(httpClientErrorException);
```
iguannago: Please, notice that if the system fails to upload the json then it throws an exception that bubbles up to the end users. I don't think we want the user to know that we fail to upload the robotic JSON to the DocStore, do we? 
I think we just need to catch any Exception and log it as an info message. 
Make sure, this method is not used somewhere else as the error handling then required there might be different

src/main/java/uk/gov/hmcts/reform/sscs/service/SubmitAppealService.java
chris-moreton: That's a mistake that I made while resolving the merge conflicts. This got picked up by the unit tests.

src/main/java/uk/gov/hmcts/reform/sscs/service/RoboticsJsonUploadService.java
chris-moreton: Yes, that's a good point. We don't want the system to report technical difficulties in this situation.

jackmaloney:
This PR has been open for so long, closing as too far out of sync with master now. The ticket should be reimplemented in Evidence Share service if it is still required.

PR #385, opened 2019-02-20
SSCS-4841 SYA Report (for reconciliation with Robot Batch Report)

srochani:


src/main/java/uk/gov/hmcts/reform/sscs/service/SubmitAppealService.java
```diff
@@ -54,7 +54,7 @@
         this.evidenceManagementService = evidenceManagementService;
     }
 
-    public void submitAppeal(SyaCaseWrapper appeal) {
+    public Long submitAppeal(SyaCaseWrapper appeal) {
```
ryandac: 👍

src/main/java/uk/gov/hmcts/reform/sscs/controller/SyaController.java
```diff
@@ -36,8 +36,9 @@
     public ResponseEntity<String> createAppeals(@RequestBody SyaCaseWrapper syaCaseWrapper) {
         log.info("Appeal with Nino - {} and benefit type {} received", syaCaseWrapper.getAppellant().getNino(),
                 syaCaseWrapper.getBenefitType().getCode());
-        submitAppealService.submitAppeal(syaCaseWrapper);
-        log.info("Appeal with Nino - {} and benefit type - {} processed successfully",
+        Long caseId = submitAppealService.submitAppeal(syaCaseWrapper);
+        log.info("Case {} with Nino - {} and benefit type - {} processed successfully",
```
ryandac: Do we still want to return a status 201 for a null caseId?

srochani: If any exception then it will return not return 201 status code and it will throw a error its in SubmitAppealService.createCaseInCcd method

ryandac: Spoke to @srochani. We will get a null caseId when it's a duplicate case. We will not create the case but for reasons unknown to us still return a 201. 

When an appeal cannot be created in ccd an exception is thrown and handled.

src/main/java/uk/gov/hmcts/reform/sscs/controller/SyaController.java
```diff
@@ -36,8 +36,9 @@
     public ResponseEntity<String> createAppeals(@RequestBody SyaCaseWrapper syaCaseWrapper) {
         log.info("Appeal with Nino - {} and benefit type {} received", syaCaseWrapper.getAppellant().getNino(),
                 syaCaseWrapper.getBenefitType().getCode());
-        submitAppealService.submitAppeal(syaCaseWrapper);
-        log.info("Appeal with Nino - {} and benefit type - {} processed successfully",
+        Long caseId = submitAppealService.submitAppeal(syaCaseWrapper);
```
iguannago: It's better to use primitives rather than boxed primities, since you are not creating new objects each time.

src/main/java/uk/gov/hmcts/reform/sscs/controller/SyaController.java
iguannago: @ryandac and @srochani , when the user submits a case that it's already in CCD (duplicate) then we return back to the user as if he/she was created that case successfully. That was agreed with Thor and Robin.

src/main/java/uk/gov/hmcts/reform/sscs/controller/SyaController.java
srochani: Yes but here its returning the caseId which is boxed primitives Long

PR #461, opened 2019-04-09
SSCS-5436 - Create PUT /drafts

ryandac:
* Made the drafts endpoint a PUT instead of a POST
* Made the endpoint update a users existing draft if it exists.
* Added a functional test to ensure there is just one draft for a user. 
   * You may need to remove ccd data if you've run this test before.
     > docker exec -ti compose_ccd-shared-database_1 psql -U postgres
     > \c ccd_data;
     > delete from case_event where event_id = 'draft';
     > delete  from case_data where state = 'draft';
  * This might need to be done on AAT if functional tests run on that environment
* The functional tests needs a ccd file change to add a _draft_ precondition for the _draft_ event


src/main/java/uk/gov/hmcts/reform/sscs/config/CitizenCcdClient.java
```diff
@@ -45,5 +49,49 @@ public CaseDetails submitForCitizen(IdamTokens idamTokens, CaseDataContent caseD
...
+
+    List<CaseDetails> searchForCitizen(IdamTokens idamTokens) {
+        List<CaseDetails> caseDetailsList =  coreCaseDataApi.searchForCitizen(
+                idamTokens.getIdamOauth2Token(),
+                idamTokens.getServiceAuthorization(),
+                idamTokens.getUserId(),
+                ccdRequestDetails.getJurisdictionId(),
+                ccdRequestDetails.getCaseTypeId(),
+                new HashMap<>()
+        );
```
ryandac: Ideally it would be good to filter on draft appeal cases, but it's difficult to tell what values are allowed in the Map for this api call.

iguannago: no need, the search will return only cases created by that specific user. And since the user is a citizen it can only create drafts, so the search will return only drafts. Actually, only one single draft.

PR #648, opened 2019-07-25
SSCS-6181 - Set Evidence Handled to "Yes" when there are no errors.

ryandac:
When running "Action Further Evidence" event, don't create a Document when the scannedDocument.document_type is "Coversheet".

The evidenceHandled should be set to “Yes if there are no errors.

if we remove the Scanned Documents, this must be set to "Yes", otherwise the case is in an incomplete state
If Scanned Document on case, evidenceHandled = "No"
if Scanned Documents were on case, but we remove them , then evidenceHandled = "Yes"

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/actionfurtherevidence/HandleEvidenceEventHandler.java
+++ b/src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/actionfurtherevidence/HandleEvidenceEventHandler.java
@@ -15,7 +15,6 @@ import java.util.ArrayList;
 import java.util.List;
 
 import lombok.extern.slf4j.Slf4j;
-import org.apache.commons.lang.StringUtils;
 import org.springframework.stereotype.Component;
 import uk.gov.hmcts.reform.sscs.ccd.callback.Callback;
 import uk.gov.hmcts.reform.sscs.ccd.callback.CallbackType;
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/actionfurtherevidence/HandleEvidenceEventHandler.java
```diff
@@ -70,17 +69,17 @@ private void buildSscsDocumentFromScan(SscsCaseData sscsCaseData) {
...
-                    documents.add(sscsDocument);
-                    sscsCaseData.setSscsDocument(documents);
-                    sscsCaseData.setEvidenceHandled(workOutEvidenceHandled(sscsCaseData.getEvidenceHandled(),
-                        sscsDocument.getValue().getDocumentType()));
+                    if (!equalsIgnoreCase(scannedDocument.getValue().getType(), COVERSHEET)) {
+                        SscsDocument sscsDocument = buildSscsDocument(sscsCaseData, scannedDocument);
+                        documents.add(sscsDocument);
+                        sscsCaseData.setSscsDocument(documents);
+                    }
+                    sscsCaseData.setEvidenceHandled("Yes");
```
iguannago: Do we want to set this to Yes for the scenario when the evidence is of Other Document type?? 
this change is changing the last AC in this ticket t https://tools.hmcts.net/jira/browse/SSCS-6026

iguannago: the BA confirmed that the AC in the 6026 is too narrow and it shouldn't be like that. Therefore, this change makes sense to me

PR #688, opened 2019-08-21
SSCS-5956 : Reissue further evidence

ryandac:


src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/reissuefurtherevidence/ReissueFurtherEvidenceAboutToSubmitHandler.java
```diff
@@ -0,0 +1,75 @@
...
+import uk.gov.hmcts.reform.sscs.ccd.presubmit.PreSubmitCallbackHandler;
+
+@Service
+public class ReissueFurtherEvidenceAboutToSubmitHandler implements PreSubmitCallbackHandler<SscsCaseData> {
+
+    private static Map<String, String> originalSenderToDocumentType = new HashMap<>();
+
+    static {
+        originalSenderToDocumentType.put(APPELLANT.getCode(), APPELLANT_EVIDENCE.getValue());
+        originalSenderToDocumentType.put(REPRESENTATIVE.getCode(), REPRESENTATIVE_EVIDENCE.getValue());
```
jackmaloney: Shouldn't it only populate this option if the case has a rep on the case?

ryandac: This is a static lookup, not based on any single appeal. Used to check what to set the document type to based on what the original sender has been selected in the mid-Event. (The about to start handler - clears out the original sender - this is because we're not going to be using the midEvent that allows the user to pick the new original sender on this ticket)

PR #730, opened 2019-09-16
Add email_to / email_from + simplification

timja:
CHG0034613

### Change description ###
Email to / email from configuration was missing in production causing defaults to be taken, this sets it up so that it's correctly pulled from vault.

I've updated chart-java to allow for templating of env variables, to reduce a lot of duplication

and moved non secret values to env vars to make debugging easier

```
[ ] Yes
[ ] No
```

### DIFF
--- a/charts/sscs-tribunals-api/values.preview.template.yaml
+++ b/charts/sscs-tribunals-api/values.preview.template.yaml
@@ -1,8 +1,10 @@
 java:
   image: ${IMAGE_NAME}
   ingressHost: ${SERVICE_FQDN}
+  aadIdentityName:
   environment:
     PDF_API_URL: "http://cmc-pdf-service-aat.service.core-compute-aat.internal"
     CORE_CASE_DATA_API_URL: "http://ccd-data-store-api-aat.service.core-compute-aat.internal"
-    IDAM_OAUTH2_REDIRECT_URL: "https://evidence-sharing-preprod.sscs.reform.hmcts.net"
     DOCUMENT_MANAGEMENT_URL: "http://dm-store-aat.service.core-compute-aat.internal"
+    IDAM_S2S_AUTH: "http://rpe-service-auth-provider-aat.service.core-compute-aat.internal"
+    IDAM_URL: https://idam-api.aat.platform.hmcts.net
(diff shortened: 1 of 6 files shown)


charts/sscs-tribunals-api/values.preview.template.yaml
```diff
@@ -1,8 +1,10 @@
 java:
   image: ${IMAGE_NAME}
   ingressHost: ${SERVICE_FQDN}
+  aadIdentityName:
   environment:
     PDF_API_URL: "http://cmc-pdf-service-aat.service.core-compute-aat.internal"
     CORE_CASE_DATA_API_URL: "http://ccd-data-store-api-aat.service.core-compute-aat.internal"
-    IDAM_OAUTH2_REDIRECT_URL: "https://evidence-sharing-preprod.sscs.reform.hmcts.net"
     DOCUMENT_MANAGEMENT_URL: "http://dm-store-aat.service.core-compute-aat.internal"
+    IDAM_S2S_AUTH: "http://rpe-service-auth-provider-aat.service.core-compute-aat.internal"
```
adusumillipraveen: We don't need these i believe? 
```suggestion
```

charts/sscs-tribunals-api/values.preview.template.yaml
```diff
@@ -1,8 +1,10 @@
...
+    IDAM_URL: https://idam-api.aat.platform.hmcts.net
```
adusumillipraveen:
```suggestion
```

timja: this needs to be here as we don't want to talk to the preview idam

charts/sscs-tribunals-api/values.preview.template.yaml
timja: we do as env would be set to preview in the preview env?

timja:
> This might be a breaking change till we bump chart version in flux. But that's fine as its not live

why? flux won't pick up a change automatically

charts/sscs-tribunals-api/values.preview.template.yaml
adusumillipraveen: It is set to aat in jenkins as we read aat vaults for preview.

PR #755, opened 2019-09-27
Update dockerfile

jackmaloney:


Dockerfile
```diff
@@ -1,15 +1,9 @@
-ARG APP_INSIGHTS_AGENT_VERSION=2.5.0-BETA.5
+ARG APP_INSIGHTS_AGENT_VERSION=2.3.1
```
timja:
```suggestion
ARG APP_INSIGHTS_AGENT_VERSION=2.5.0
```

timja: should also update this:
https://github.com/hmcts/sscs-tribunals-case-api/blob/master/build.gradle#L196
to 5.1.0 not that there's anything different in there, don't think there's any code changes but good to get off RC versions

PR #769, opened 2019-10-04
Sscs 6730 fix null

chrisdavecm:
https://tools.hmcts.net/jira/browse/SSCS-6730

### Change description ###

```
[ ] Yes
[x] No
```


jackmaloney:
But do we know why this is throwing a NPE. A case should always have a benefit code, so maybe this PR is just masking some other issue?

PR #773, opened 2019-10-08
SSCS-6664 Handle issue decision event

jackmaloney:
https://tools.hmcts.net/jira/browse/SSCS-6664


src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/decisionissued/DecisionIssuedAboutToSubmitHandler.java
```diff
@@ -0,0 +1,78 @@
...
+                    .documentType(DocumentType.DECISION_NOTICE.getValue())
+                    .build();
+
+            caseData.setSscsInterlocDecisionDocument(document);
+        }
+        saveToHistory(caseData);
+        clearTransientFields(caseData);
+
+        if (caseData.getDecisionType() != null && caseData.getDecisionType().equals(STRIKE_OUT.getValue())) {
+            caseData.setOutcome("nonCompliantAppealStruckout");
```
ryandac: These magic string do look a bit odd. Not sure what to do about it - probably leave it as it is. It might be better served as an enum like the decisionType would be, but that would involve the headache of changing sscs-common code.

PR #789, opened 2019-10-21
Feature/sscs 6759 send to admin

iguannago:
https://tools.hmcts.net/jira/browse/SSCS-6759


ryandac:
Don't forget to remove the output.pdf file 😃

PR #808, opened 2019-10-25
Found a library that converts tiff images. 

ryandac:
Found a library that converts tiff images.
Tiff's are not supported by default, but will be in Java 9.


jackmaloney:
Approved but guess you might want to remove the output.pdf file

ryandac:
Closing this PR.

PR #812, opened 2019-10-28
Task/sscs 6808

iguannago:
SSCS-6808


charts/sscs-tribunals-api/values.yaml
```diff
@@ -50,4 +50,5 @@ java:
     DOCUMENT_MANAGEMENT_URL: "http://dm-store-{{ .Values.global.environment }}.service.core-compute-{{ .Values.global.environment }}.internal"
     IDAM_S2S_AUTH: "http://rpe-service-auth-provider-{{ .Values.global.environment }}.service.core-compute-{{ .Values.global.environment }}.internal"
     IDAM_API_URL: https://idam-api.{{ .Values.global.environment }}.platform.hmcts.net
-    ISSUE_FURTHER_EVIDENCE_ENABLED: "false"
\ No newline at end of file
+    ISSUE_FURTHER_EVIDENCE_ENABLED: "false"
+    READY_TO_LIST_OFFICES: ""
```
jackmaloney: I think the proper way to do this is to put `DWP PIP (3)` here. And then in cnp-flux-config set the prod value for this config value to empty. And then merge the prod config before you merge this PR

PR #832, opened 2019-11-06
Feature/sscs 5955 action further evidence intermediate interloc review

iguannago:
https://tools.hmcts.net/jira/browse/SSCS-5955


src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/actionfurtherevidence/ActionFurtherEvidenceSubmittedCallbackHandler.java
```diff
@@ -32,31 +40,32 @@ public ActionFurtherEvidenceSubmittedCallbackHandler(CcdService ccdService, Idam
...
         requireNonNull(callback, "callback must not be null");
         requireNonNull(callbackType, "callbackType must not be null");
+        DynamicList furtherEvidenceAction = callback.getCaseDetails().getCaseData().getFurtherEvidenceAction();
         return callbackType.equals(CallbackType.SUBMITTED)
-                && callback.getEvent().equals(EventType.ACTION_FURTHER_EVIDENCE)
-                && (isInformationReceivedForInterlocType(callback.getCaseDetails().getCaseData().getFurtherEvidenceAction(), INFORMATION_RECEIVED_FOR_INTERLOC_JUDGE)
-                || isInformationReceivedForInterlocType(callback.getCaseDetails().getCaseData().getFurtherEvidenceAction(), INFORMATION_RECEIVED_FOR_INTERLOC_TCW)
-                || isIssueToAllParties(callback.getCaseDetails().getCaseData().getFurtherEvidenceAction()));
+            && callback.getEvent().equals(EventType.ACTION_FURTHER_EVIDENCE)
+            && furtherEvidenceActionOptionValidation(furtherEvidenceAction);
```
jackmaloney: What is the point of this `furtherEvidenceActionOptionValidation`? Doesn't that include every option and so it will always be true?

iguannago: If we assume that this submitted callback handler is always called after the about to submit callback, then yes, that furtherEvidenceActionOptionValidation field SHOULD always contain a valid option. However, that is the assumption or we can also call it, the happy path, ritht? 

However, what if someone adds more options intentionally or unintentionally (by mistake)? This check is ensuring the data coming is valid or as expected before carrying on

PR #864, opened 2019-11-27
SSCS-6950

iguannago:
https://tools.hmcts.net/jira/browse/SSCS-6950


iguannago:
no needed anymore since it's in master already

PR #881, opened 2019-12-17
Feature/sscs 6587

iguannago:
https://tools.hmcts.net/jira/browse/SSCS-6587


src/main/java/uk/gov/hmcts/reform/sscs/service/SubmitAppealService.java
```diff
@@ -122,8 +118,10 @@ private void postCreateCaseInCcdProcess(SscsCaseData caseData,
         }
     }
 
-    SscsCaseData prepareCaseForCcd(SyaCaseWrapper appeal, String postcode) {
-        RegionalProcessingCenter rpc = regionalProcessingCenterService.getByPostcode(postcode);
+    SscsCaseData convertAppealToSscsCaseData(SyaCaseWrapper appeal) {
+        String firstHalfOfPostcode = regionalProcessingCenterService
+            .getFirstHalfOfPostcode(appeal.getContactDetails().getPostCode());
```
srochani: Some UK postcode has 4 first half. Examples of postcodes are "SW1W 0NY", "PO16 7GZ"

iguannago: I haven't changed that. I just only moved the getFirstHalfOfPostcode method to a different method where it is actually needed and therefore it makes more sense to for this method to live. 
I haven't changed at all the way we handle the postcodes etc.

PR #938, opened 2020-02-19
Add log statement before hitting bundling service

jackmaloney:


src/main/resources/config/application-local.properties
```diff
@@ -2,12 +2,12 @@
 spring.profiles=local
 
 # LOGGING
-logging.level.root=DEBUG
-logging.level.org.springframework.web=${LOG_LEVEL_SPRING_WEB:debug}
-logging.level.uk.gov.hmcts.reform.sscs=${LOG_LEVEL_SSCS:debug}
-logging.level.uk.gov.hmcts.reform.sscs.idam.IdamApiClient=DEBUG
-logging.level.uk.gov.hmcts.reform.ccd.client.CoreCaseDataApi=DEBUG
-logging.level.uk.gov.hmcts.reform.ccd.document.EvidenceMetadataDownloadClientApi=DEBUG
+logging.level.root=INFO
```
iguannago: this is a local profile to turn ON the debug mode and only for development and local usage. If you turn OFF the debug mode then it does not make sense to keep this profile.

jackmaloney: This profile now has the bundling url so needs to always be run locally like this.

PR #946, opened 2020-02-24
SSCS-7123 Prevent DD case code

jackmaloney:
https://tools.hmcts.net/jira/browse/SSCS-7123


srochani:
You have to change in CaseCodeService class for default issue code from DD

PR #964, opened 2020-03-02
Sscs 7001 ignore coversheet

jackmaloney:
https://tools.hmcts.net/jira/browse/SSCS-7001

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/furtherevidence/actionfurtherevidence/ActionFurtherEvidenceAboutToSubmitHandler.java
+++ b/src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/furtherevidence/actionfurtherevidence/ActionFurtherEvidenceAboutToSubmitHandler.java
@@ -64,7 +64,7 @@ public class ActionFurtherEvidenceAboutToSubmitHandler implements PreSubmitCallb
             sscsCaseData.setDwpFurtherEvidenceStates(FURTHER_EVIDENCE_RECEIVED);
         }
 
-        buildSscsDocumentFromScan(sscsCaseData, caseDetails.getState());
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/furtherevidence/actionfurtherevidence/ActionFurtherEvidenceAboutToSubmitHandler.java
```diff
@@ -117,6 +111,24 @@ private void buildSscsDocumentFromScan(SscsCaseData sscsCaseData, State caseStat
...
+            preSubmitCallbackResponse.addError("No document URL so could not process");
+        }
+
+        if (scannedDocument.getValue().getFileName() == null) {
+            preSubmitCallbackResponse.addError("No document file name so could not process");
+        }
+
+        if (equalsIgnoreCase(scannedDocument.getValue().getType(), COVERSHEET)) {
+            if (!ignoreWarnings) {
+                preSubmitCallbackResponse.addWarning("Coversheet will be ignored, are you happy to proceed?");
```
srochani: Message doesn't look right, as if ignoreWarning is false then message says conversheet will be ignored

jackmaloney: Yeh that is correct. The way CCD works is the first time you press submit, the 'ignoreWarnings' flag in the payload is false. If we return a warning in our response to the user, they have the option of saying yes/no to ignore warning. Then, we receive the second payload, which will have the 'ignoreWarnings' as true, if they selected yes. In this case, we don't want to return another warning as they have already confirmed they are happy to proceed.

PR #1092, opened 2020-06-10
SSCS-7623 Handle empty hearings for generating decision notices

jackmaloney:
https://tools.hmcts.net/jira/browse/SSCS-7623

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/writefinaldecision/WriteFinalDecisionMidEventHandler.java
+++ b/src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/writefinaldecision/WriteFinalDecisionMidEventHandler.java
@@ -123,7 +123,7 @@ public class WriteFinalDecisionMidEventHandler extends IssueDocumentHandler impl
             }
         } else {
             builder.heldOn(LocalDate.now());
-            builder.heldAt("In-chambers");
+            builder.heldAt("In chambers");
         }
 
         Outcome outcome = decisionNoticeOutcomeService.determineOutcome(caseData);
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/writefinaldecision/WriteFinalDecisionMidEventHandler.java
```diff
@@ -121,6 +121,9 @@ protected DirectionOrDecisionIssuedTemplateBody createPayload(SscsCaseData caseD
                     builder.heldAt(finalHearing.getValue().getVenue().getName());
                 }
             }
+        } else {
```
michaellavelle, line 124: How do we want to handle the conditions here where either i)  the first hearing in the list is null, or a hearing that has no value ( perhaps that can't happen?), or ii) the first hearing in the list has no hearing date,  iii)  the first hearing in the list has no venue, or iv)  the first hearing in the list has a venue, but no name ( again, perhaps that can't happen?) ?     Do we want to default the details in those cases,  or should an IllegalStateException be thrown ?

jackmaloney: I was thinking the same thing and decided to leave the code in that throws an 'IllegalStateException' rather than defaulting - a hearing should never be null for any of those details. If it is then we should probably investigate and find out how/why this is happening. Of course, if we were getting lots of these and there was a genuine reason for it then we would need to start defaulting it.

michaellavelle: Agreed - sounds like a good strategy - we have the validation code in there as you say.  Let's go with this,  unless Rosie suggests otherwise.

PR #1104, opened 2020-06-15
SSCS-7415 - CCD UI - Strikeout cases - event

ryandac:
# [SSCS-7415](https://tools.hmcts.net/jira/browse/SSCS-7415) CCD UI - Strikeout cases - event

At the moment when judge(s)/TCW(s) issue a notice, they may say that if there is no response in 14 days, then the appeal will be struck out. In terms of business terminology this is an automatic strike out – meaning that it doesn’t need to go back to a Judge/TCW to be struck out and admins should also be able to do this. When this happens, a letter is sent to all parties, the due date is removed, the case is flagged to DWP and the case should be moved to dormant, this ticket will be covering all aspects of this.


src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/struckout/StruckOutAboutToSubmitHandler.java
```diff
@@ -0,0 +1,47 @@
...
+            throw new IllegalStateException("Cannot handle callback.");
+        }
+        final CaseDetails<SscsCaseData> caseDetails = callback.getCaseDetails();
+        final SscsCaseData sscsCaseData = caseDetails.getCaseData();
+
+        log.info(String.format("Handling struckOut event for caseId %s", sscsCaseData.getCcdCaseId()));
+
+        sscsCaseData.setInterlocReviewState(null);
+        sscsCaseData.setDirectionDueDate(null);
+        sscsCaseData.setHmctsDwpState("strikeOutActioned");
```
srochani, line 41: Can you please add "strikeOutActioned" as constant variable or state also do we have to set the DwpState here? e.g caseData.setDwpState(DwpState.STRUCK_OUT.getId());

PR #1147, opened 2020-06-30
SSCS-7730 - CCD UI – Have directionIssued validate prior to submitting the event

ryandac:
At the moment we are getting exceptions when users are doing the directionIssued event and they have selected “appealToProcceed” for the directionTypeDl, when the event is submitted this, then makes a call for the “appealToProceed” event to trigger. 

The exception occurs when the event is submitted and there is some data which causes the bulk scan validation to fail, all of which happens asynchronously, so the user is unaware that their case hasn’t progressed.

As part of the investigation into the issue that was seen against SSCS-7695, we want to hit the bulk scan OCR validation end point during the about to submit callback for the direction issued event, if they have selected the “appeal to proceed” direction type, if the validation were to then fail, then the error should be displayed back to the user.

The errors from the screenshot below are propagated from sscs-bulk-scan.

Screenshot 2020-06-30 at 10 14 06

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/directionissued/DirectionIssuedAboutToSubmitHandler.java
+++ b/src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/directionissued/DirectionIssuedAboutToSubmitHandler.java
@@ -1,5 +1,6 @@
 package uk.gov.hmcts.reform.sscs.ccd.presubmit.directionissued;
 
+import static org.apache.commons.lang3.StringUtils.trimToEmpty;
 import static uk.gov.hmcts.reform.sscs.ccd.presubmit.InterlocReviewState.AWAITING_ADMIN_ACTION;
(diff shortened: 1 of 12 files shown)


src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/directionissued/DirectionIssuedAboutToSubmitHandler.java
```diff
@@ -88,6 +99,12 @@ public boolean canHandle(CallbackType callbackType, Callback<SscsCaseData> callb
         caseData.setTimeExtensionRequested("No");
 
         PreSubmitCallbackResponse<SscsCaseData> sscsCaseDataPreSubmitCallbackResponse = new PreSubmitCallbackResponse<>(caseData);
+
+        if (caseDetails.getState().equals(State.INTERLOCUTORY_REVIEW_STATE) && caseData.getDirectionTypeDl() != null && StringUtils.equals(DirectionType.APPEAL_TO_PROCEED.toString(), caseData.getDirectionTypeDl().getValue().getCode())) {
+            // do something
```
danhamilton, line 104: comment needed?

PR #1216, opened 2020-07-29
SSCS-7854 Adding workaround for RDM-8200 to set the generated dates on the AboutToSubmit handlers

michaellavelle:

https://tools.hmcts.net/jira/browse/SSCS-7854

### Change description ###

*   Adding workaround for RDM-8200 to set the generated dates on the AboutToSubmit handlers.  Due to the generated dates being a hidden field, these values which are already set are being nulled out.   This is a workaround for that issue

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/writefinaldecision/WriteFinalDecisionAboutToSubmitHandler.java
+++ b/src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/writefinaldecision/WriteFinalDecisionAboutToSubmitHandler.java
@@ -3,6 +3,7 @@ package uk.gov.hmcts.reform.sscs.ccd.presubmit.writefinaldecision;
 import static org.apache.commons.collections4.CollectionUtils.emptyIfNull;
 import static uk.gov.hmcts.reform.sscs.ccd.callback.DocumentType.DRAFT_DECISION_NOTICE;
 
+import java.time.LocalDate;
 import java.util.*;
 import java.util.stream.Collectors;
 import lombok.extern.slf4j.Slf4j;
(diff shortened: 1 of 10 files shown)


src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/writefinaldecision/WriteFinalDecisionAboutToSubmitHandler.java
```diff
@@ -51,7 +51,9 @@ public boolean canHandle(CallbackType callbackType, Callback<SscsCaseData> callb
         // on the final submission from CCD, so we need to reset it here
         // See https://tools.hmcts.net/jira/browse/RDM-8200
         // This is a temporary workaround for this issue.
-        sscsCaseData.setWriteFinalDecisionGeneratedDate(LocalDate.now().toString());
+        if (sscsCaseData.getWriteFinalDecisionGeneratedDate() == null) {
```
jackmaloney, line 54: Im not sure if this is going to work for the following scenario:
- Final decision generated on say 1st August. Document generated date says '1st August' and the writeFinalDecisionGeneratedDate field is saved with '1st August'
- Peer reviewed offline and they need to make a change the following day. The document generated date for the preview says '2nd August'. 
- The Final decision is then submitted. Due to line 54 only overwriting the writeFinalDecisionGeneratedDate if it is null, then it will leave it as the 1st August (even though the preview says the 2nd August. 
- When the document is issued, it will use the date that is saved in CCD which is the 1st August - this is not correct.

michaellavelle: Thanks for going through this Jack.   We've now demonstrated that this is indeed the case - I've removed the conditional null check

PR #1291, opened 2020-08-20
SSCS-8004 - Bundling – Tribunals – Move the bundle endpoint call from the AboutToSubmit handler to Submitted

ryandac:
SSCS-8004 - Bundling – Tribunals – Move the bundle endpoint call from the AboutToSubmit handler to Submitted


src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/createbundle/CreateBundleAboutToStartHandler.java
```diff
@@ -4,8 +4,6 @@
 import static uk.gov.hmcts.reform.sscs.model.AppConstants.DWP_DOCUMENT_EVIDENCE_FILENAME_PREFIX;
```
srochani, line 4: Do we still require this about to Submit callback?

ryandac: Yes. It sends an error if mandatory files are missing. It also updates the case. 
It's too late to update the case on the Submitted - you can do it - but you'd have to call an update event.

The call to `serviceRequestExecutor.post(callback, bundleUrl + CREATE_BUNDLE_ENDPOINT);` could also return an error or warning (maybe rare that it could occur) that we will miss on the ccd ui as it's on the submitted event. Perhaps I could add some logging if that call returns any errors or warnings

srochani (approved):

ryandac:
not needed

PR #1588, opened 2020-11-25
SSCS-7580 Ensure that we validate that reg35 must be set to No for Scenario 5

michaellavelle:

https://tools.hmcts.net/jira/browse/SSCS-7580

### Change description ###

Scenario 5 stipulates that Regulation 35 is set to no,  but if the user selected yes, the same Scenario was selected and the decision notice displayed, but with incorrect content.

This PR validates that Regulation 35 is set to Yes for Scenario 5 and displays an error message is so. 
Note :  Since creating this PR a new Scenario 12 has been added which covers the case that this validates for - making it a valid scenario but with it's own content.   We may want to merge this first,  and then add Scenario 12 later,  or alternatively wait for Scenario 12 to be implemented - in which case this PR won't be needed

```
[ ] Yes
[x] No
```


michaellavelle:
Superseded by https://github.com/hmcts/sscs-tribunals-case-api/pull/1613

PR #1591, opened 2020-11-27
SSCS-8083  - Scenario 8 and 9

ryandac:
SSCS-8083: Scenario 8 and 9


chrisdavecm (commented):
Can we delete the old tests for 8 and 9?

chrisdavecm (approved):

PR #1651, opened 2020-12-09
SSCS-8503 -  Populate venue field with the venue a case is expected to be managed from at creation	

ryandac:
## SSCS-8503 -  Populate venue field with the venue a case is expected to be managed from at creation


src/main/java/uk/gov/hmcts/reform/sscs/service/SubmitAppealService.java
```diff
@@ -145,8 +147,8 @@ private void postCreateCaseInCcdProcess(SscsCaseData caseData,
 
     SscsCaseData convertAppealToSscsCaseData(SyaCaseWrapper appeal) {
 
-        String firstHalfOfPostcode = regionalProcessingCenterService
-            .getFirstHalfOfPostcode(appeal.getContactDetails().getPostCode());
+        String postCode = appeal.getContactDetails().getPostCode();
+        String firstHalfOfPostcode = getFirstHalfOfPostcode(postCode);
         RegionalProcessingCenter rpc = regionalProcessingCenterService.getByPostcode(firstHalfOfPostcode);
```
rasikasri, line 152: Ryan, As you have done some refactoring as well, Isn't it nice to let getByPostcode method to get the first half of the post code? As both the methods are in RegionalProcessingCenterService and assuming getByPostcode would not work differently if you pass the full postcode.

rasikasri (approved):

src/main/java/uk/gov/hmcts/reform/sscs/service/SubmitAppealService.java
ryandac: That method is in sscs-common, it already does concatenation of the string postcode. It was not in my remit to change it. If I was to I would make it less mad and write lots of unit tests to handle full and partial postcodes. Your welcome to change it, I didn't because it didn't affect populating the venue.

rasikasri: Agree with you, I added this comment as you had done some refactoring and in case if you had missed this point. I'll mark this as resolved.

src/test/java/uk/gov/hmcts/reform/sscs/service/SubmitAppealServiceTest.java
```diff
@@ -518,4 +496,24 @@ public void getMatchedCases() {
...
+    @Parameters({
+            "PIP, n1w1 wal, Birmingham",
+            "ESA, n1w1 wal, Birmingham",
+            "UC, n1w1 wal, Birmingham",
+            "PIP, NN85 1ss, Northampton",
+            "ESA, NN85 1ss, Northampton",
+            "UC, NN85 1ss, Northampton"
+
+    })
+    public void shouldSetProcessingVenueBasedOnBenefitTypeAndPostCode(String benefitCode, String postcode, String expectedVenue) {
```
jackmaloney, line 510: Could you add a test to cover the appointee route please?

ryandac: added

PR #1717, opened 2021-01-20
Sscs 8431 remove feature flags

jackmaloney:
https://tools.hmcts.net/jira/browse/SSCS-8431


srochani (approved):
I think from charts yml also it needs to be removed

PR #1930, opened 2021-03-30
Revert "SSCS-8827 Single case from draft case"

jackmaloney:
Reverts hmcts/sscs-tribunals-case-api#1870


src/main/java/uk/gov/hmcts/reform/sscs/config/CitizenCcdService.java
```diff
@@ -89,6 +90,22 @@ public SaveCaseResult createDraft(SscsCaseData caseData, IdamTokens idamTokens)
                 .build();
     }
 
+    public Optional<SscsCaseDetails> draftArchivedFirst(SscsCaseData caseData,
```
srochani, line 93: I think we have removed the draft archived right.. now we are not archiving just creating the case from draft

PR #2078, opened 2021-05-27
sscs 9105 return booleans rather than Yes/No

chrisdavecm:
https://tools.hmcts.net/jira/browse/SSCS-9105

### Change description ###
Send true and false out to javascript rather than yes, no or undefined

```
[ ] Yes
[x] No
```


divisathyan (approved):
LGTM

PR #2111, opened 2021-06-10
SSCS-9294 - Enable Pact tests on path to live

ryandac:
SSCS-9294 - Enable Pact tests on path to live

This PR has verified that the pact tests run and are published on a PR commit, since we only want it to be run on a commit to master it has been removed.


jhart-dev (approved):

ryandac:
no longer needed

PR #2112, opened 2021-06-10
SSCS-9292 Show error when rip1 uploaded but no av evidence

jackmaloney:
https://tools.hmcts.net/jira/browse/SSCS-9292


divisathyan (approved):
LGTM

PR #2523, opened 2022-01-12
SSCS-9842:SSCS5 Update other party data event

divisathyan:
https://tools.hmcts.net/jira/browse/SSCS-9842

```
[ ] Yes
[x ] No
```


src/main/java/uk/gov/hmcts/reform/sscs/util/OtherPartyDataUtil.java
```diff
@@ -225,4 +225,32 @@ public static boolean hasNewOtherPartyAdded(List<CcdValue<OtherParty>> before, L
         }
         return false;
     }
+
+    public static void validateOtherPartyForSscs5Case(SscsCaseData sscsCaseData) {
```
ryandac, line 229: bit strange that this method is called validate, but it doesn't validate anything

ryandac (approved):

PR #2922, opened 2022-09-01
SSCS-10790:Adding hearing date to excluded date list

EliandroGV1:

https://tools.hmcts.net/jira/browse/SSCS-10790

### Change description ###
Current hearing date needs to be added to the current excluded date list

```
[ ] Yes
[X] No
```


src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/postponementrequest/PostponementRequestAboutToStartHandler.java
```diff
@@ -50,7 +51,12 @@ public PreSubmitCallbackResponse<SscsCaseData> handle(CallbackType callbackType,
     }
 
     private void setPostponementRequest(Hearing hearing, SscsCaseData sscsCaseData) {
-        sscsCaseData.getPostponementRequest().setPostponementRequestHearingDateAndTime(hearing.getValue().getHearingDateTime().toString());
+        if (HearingRoute.GAPS.equals(sscsCaseData.getRegionalProcessingCenter().getHearingRoute())) {
```
TeaWithLucas, line 54: I think you should be using the `getScheulingAndListingFields().getHearingRoute()`

src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/postponementrequest/PostponementRequestAboutToStartHandler.java
```diff
@@ -50,7 +51,12 @@ public PreSubmitCallbackResponse<SscsCaseData> handle(CallbackType callbackType,
...
+            sscsCaseData.getPostponementRequest().setPostponementRequestHearingDateAndTime(hearing.getValue().getHearingDateTime().toString());
+
+        } else if (HearingRoute.LIST_ASSIST.equals(sscsCaseData.getRegionalProcessingCenter().getHearingRoute())) {
```
TeaWithLucas, line 57: I think you should be using the `getScheulingAndListingFields().getHearingRoute()`

src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/actionpostponementrequest/ActionPostponementRequestAboutToSubmitHandler.java
```diff
@@ -108,7 +118,12 @@ private void grantPostponement(SscsCaseData sscsCaseData, PostponementRequest po
...
         final Optional<Hearing> optionalHearing = emptyIfNull(sscsCaseData.getHearings()).stream()
-                .filter(h -> h.getValue().getHearingDateTime().isAfter(LocalDateTime.now()))
+                .filter(h -> {
+                    if (!isNull(h.getValue().getStart())) {
+                        return LocalDateTime.now().isBefore(h.getValue().getStart());
+                    }
+                    return false;
+                })
                 .distinct()
                 .findFirst();
```
TeaWithLucas, line 129: There is no need for this, just use:
`Hearing hearing = sscsCaseData.getLatestHearing();
footerService.setHearingDateAsExcludeDate(hearing, sscsCaseData)`

src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/postponementrequest/PostponementRequestAboutToStartHandler.java
```diff
@@ -32,16 +32,17 @@ public PreSubmitCallbackResponse<SscsCaseData> handle(CallbackType callbackType,
...
         final PreSubmitCallbackResponse<SscsCaseData> response = new PreSubmitCallbackResponse<>(sscsCaseData);
 
         if (HearingRoute.GAPS.equals(sscsCaseData.getRegionalProcessingCenter().getHearingRoute())) {
             response.addError(POSTPONEMENTS_NOT_POSSIBLE_FOR_GAPS);
+            return response;
         }
+        final Optional<Hearing> optionalHearing = emptyIfNull(sscsCaseData.getHearings()).stream()
+                .filter(h -> h.getValue().getStart().isAfter(LocalDateTime.now()))
+                .distinct()
+                .findFirst();
```
TeaWithLucas, line 44: I would suggest just doing:
`Hearing hearing = sscsCaseData.getLatestHearing();
setPostponementRequest(hearing, sscsCaseData)`
instead

src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/actionpostponementrequest/ActionPostponementRequestAboutToSubmitHandler.java
```diff
@@ -31,6 +32,15 @@
...
 
+
+
+
+
+
+
+
+
+
```
TeaWithLucas, line 43: Remove these newlines

build.gradle
```diff
@@ -584,7 +584,7 @@ bootWithCCD {
     dependsOn generateXls
 
     //Uncomment to launch with debug
-    //jvmArgs = ['-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=5006']
+    jvmArgs = ['-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=5006']
```
TeaWithLucas, line 587: remember to not commit local debug stuff

src/main/resources/config/application.properties
```diff
@@ -122,7 +122,7 @@ feature.postponements.enabled: ${POSTPONEMENTS_FEATURE:false}
 feature.sscs2.enabled: ${SSCS2_FEATURE:false}
 feature.work-allocation.enabled: ${WORK_ALLOCATION_FEATURE:false}
 feature.gaps-switchover.enabled: ${GAPS_SWITCHOVER_FEATURE:false}
-feature.snl.enabled: ${SCHEDULE_LISTING_FEATURE:false}
```
TeaWithLucas, line 125: remember to not commit local debug stuff

TeaWithLucas (changes requested):

src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/postponementrequest/PostponementRequestAboutToStartHandler.java
EliandroGV1: done

src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/postponementrequest/PostponementRequestAboutToStartHandler.java
EliandroGV1: I have looked at this have seen that getLatestHearing() sorts it by hearingid (which may not be created in order) , then datetime etc but we only care about date which i think the code i pushed does : i will change it to :
```
final Optional<Hearing> optionalHearing = emptyIfNull(sscsCaseData.getHearings()).stream()
        .filter(hearing -> LocalDateTime.now().isBefore(hearing.getValue().getStart()))
        .distinct()
        .findFirst();
```

which is cleaner as there is no need to check for nullable for a postponement

src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/actionpostponementrequest/ActionPostponementRequestAboutToSubmitHandler.java
EliandroGV1: have looked at this have seen that getLatestHearing() sorts it by hearingid (which may not be created in order) , then datetime etc but we only care about date which i think the code i pushed does

filippo-possenti-version1: Agreed. The two solutions are not equivalent. Furthermore we found that `getLatestHearing` currently has side-effects so we don't want to rely on it anyway.

We'll post a small simplification of the current code, but it won't use `getLatestHearing` and that's intended.

src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/postponementrequest/PostponementRequestAboutToStartHandler.java
filippo-possenti-version1: Agreed. `getLatestHearing` is not equivalent to Eliandro's work and therefore cannot be used.

src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/actionpostponementrequest/ActionPostponementRequestAboutToSubmitHandler.java
```diff
@@ -108,7 +108,7 @@ private void grantPostponement(SscsCaseData sscsCaseData, PostponementRequest po
 
     private void setHearingDateToExcludedDate(SscsCaseData sscsCaseData, PreSubmitCallbackResponse<SscsCaseData> response) {
         final Optional<Hearing> optionalHearing = emptyIfNull(sscsCaseData.getHearings()).stream()
-                .filter(h -> h.getValue().getHearingDateTime().isAfter(LocalDateTime.now()))
+                .filter(hearing -> LocalDateTime.now().isBefore(hearing.getValue().getStart()))
```
ss-evoco, line 111: Wont this give us all the hearings that have already expired / on previous days ?

src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/postponementrequest/PostponementRequestAboutToStartHandler.java
```diff
@@ -32,25 +32,31 @@ public PreSubmitCallbackResponse<SscsCaseData> handle(CallbackType callbackType,
...
+
         optionalHearing.ifPresentOrElse(hearing -> setPostponementRequest(hearing, sscsCaseData),
                 () -> response.addError("There are no hearing to postpone"));
 
         return response;
     }
 
     private void setPostponementRequest(Hearing hearing, SscsCaseData sscsCaseData) {
```
ss-evoco, line 53: Formatting seems to have gone a bit awry in this method.

src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/postponementrequest/PostponementRequestAboutToStartHandler.java
```diff
@@ -32,25 +32,31 @@ public PreSubmitCallbackResponse<SscsCaseData> handle(CallbackType callbackType,
...
-        sscsCaseData.getPostponementRequest().setPostponementRequestHearingDateAndTime(hearing.getValue().getHearingDateTime().toString());
+        if (HearingRoute.GAPS.equals(sscsCaseData.getSchedulingAndListingFields().getHearingRoute())) {
```
ss-evoco, line 54: There are no postponements of GAPS cases. This scenario should throw an error like above.

ss-evoco (changes requested):

src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/postponementrequest/PostponementRequestAboutToStartHandler.java
filippo-possenti-version1: If you mean the indentation, that's what the IDE proposed.

src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/actionpostponementrequest/ActionPostponementRequestAboutToSubmitHandler.java
filippo-possenti-version1: From what I understand this was done on purpose, but I'll take a closer look as to the rationale behind it and get back to you.

src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/postponementrequest/PostponementRequestAboutToStartHandler.java
filippo-possenti-version1: There is a dedicated test checking that a postponement request from GAPS throws an error and it does indeed throw an error by virtue of how the "handle" method is implemented (line 39).

If anything, this "if" should not exist at all. I'll make that change.

src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/postponementrequest/PostponementRequestAboutToStartHandler.java
TeaWithLucas: It orders by hearing date, then by hearing id after. For this we want the current hearing, which is the latest, I recommend using getLatestHearing so that we have consistency on this with all our S&L work.

src/test/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/actionpostponementrequest/ActionPostponementRequestAboutToSubmitHandlerTest.java
```diff
@@ -186,6 +188,7 @@ public void givenAGrantedPostponementAndNotListable_thenClearReviewStateAndRefer
 
         sscsCaseData.setHearings(List.of(Hearing.builder().value(HearingDetails.builder()
                 .hearingDate(LocalDate.now().plusDays(1).toString())
+                        .start(LocalDateTime.now().plusDays(1))
```
TeaWithLucas, line 191: bad indent

src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/actionpostponementrequest/ActionPostponementRequestAboutToSubmitHandler.java
TeaWithLucas: Again, I think we should be using getLatestHearing so that we have consistency on this with all our S&L work.

TeaWithLucas (changes requested):

src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/postponementrequest/PostponementRequestAboutToStartHandler.java
TeaWithLucas: I aggree with Sean on this, there is a isSandL method in utilities I used it for my pr, so if(!isSandL(caseData) { addError("Case is not list assist")
there should also be a feature flag for SandL in the canHandle method

TeaWithLucas (dismissed):

ss-evoco (dismissed):

amoslv1 (approved):

src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/postponementrequest/PostponementRequestAboutToStartHandler.java
```diff
@@ -1,56 +1,65 @@
...
 
-        if (HearingRoute.GAPS.equals(sscsCaseData.getRegionalProcessingCenter().getHearingRoute())) {
-            response.addError(POSTPONEMENTS_NOT_POSSIBLE_FOR_GAPS);
+        if (!SscsUtil.isSAndLCase(sscsCaseData)) {
+            response.addError(NOT_LIST_ASSIST_CASE_ERROR);
+            return response;
+        }
+
+        Hearing hearing = sscsCaseData.getLatestHearing();
+        if (isNull(hearing)) {
```
filippo-possenti-version1, line 51: getLatestHearings is known to be a method with side effects. Plus it does additional sorting that may affect performance.

Recommend replacing the two lines with:
```
if (ObjectUtils.isEmpty(sscsCaseData.getHearings()) {
```

src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/postponementrequest/PostponementRequestAboutToStartHandler.java
```diff
@@ -1,56 +1,65 @@
...
-        final Optional<Hearing> optionalHearing = emptyIfNull(sscsCaseData.getHearings()).stream()
-                .filter(h -> h.getValue().getHearingDateTime().isAfter(LocalDateTime.now()))
-                .distinct()
-                .findFirst();
-
         final PreSubmitCallbackResponse<SscsCaseData> response = new PreSubmitCallbackResponse<>(sscsCaseData);
 
-        if (HearingRoute.GAPS.equals(sscsCaseData.getRegionalProcessingCenter().getHearingRoute())) {
-            response.addError(POSTPONEMENTS_NOT_POSSIBLE_FOR_GAPS);
+        if (!SscsUtil.isSAndLCase(sscsCaseData)) {
```
filippo-possenti-version1, line 45: Semantically different.
Before it was giving an error if the case "was not GAPS". Now it gives an error if the case "is not S&L".

Also this code may require a feature flag, which in turn will need to be crafted carefully as it might end up removing the error the previously had if the case was not GAPS.

- Recommend to restore the previous if statement unless there's a specific reason
- Recommend to evaluate whether this code requires a feature flag

src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/postponementrequest/PostponementRequestAboutToStartHandler.java
```diff
@@ -1,56 +1,65 @@
...
+            response.addError(NOT_LIST_ASSIST_CASE_ERROR);
```
filippo-possenti-version1, line 46: The error code changed. The two errors emphasize the different semantics of the if statement.
The previous one stated that postponements are not possible for GAPS, the new one states that only list assist cases can have postponements.

- Recommend to check whether there were requirements around the error phrasing or alternatively restore the previous phrasing

src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/postponementrequest/PostponementRequestAboutToStartHandler.java
```diff
@@ -1,56 +1,65 @@
...
     public boolean canHandle(CallbackType callbackType, Callback<SscsCaseData> callback) {
         requireNonNull(callback, "callback must not be null");
         requireNonNull(callbackType, "callbacktype must not be null");
 
-        return callbackType.equals(CallbackType.ABOUT_TO_START)
-                && callback.getEvent() == EventType.POSTPONEMENT_REQUEST
-                && callback.getCaseDetails() != null;
+        return callbackType == CallbackType.ABOUT_TO_START
+            && callback.getEvent() == EventType.POSTPONEMENT_REQUEST
+            && isScheduleListingEnabled;
```
filippo-possenti-version1, line 35: Why was the getCaseDetails() != null condition removed? Won't it affect the handle method potentially opening to NullPointerException being thrown in some circumstances?

src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/actionpostponementrequest/ActionPostponementRequestAboutToSubmitHandler.java
```diff
@@ -118,7 +110,7 @@ private void refusePostponement(SscsCaseData sscsCaseData) {
 
     private void grantPostponement(SscsCaseData sscsCaseData, PreSubmitCallbackResponse<SscsCaseData> response) {
         cancelHearing(sscsCaseData);
-        setHearingDateToExcludedDate(sscsCaseData, response);
+        postponementRequestService.addCurrentHearingToExcludeDates(response);
```
filippo-possenti-version1, line 113: This method is semantically different from setHearingDateToExcludedDate.

In particular, the selection of the hearing date in the previous one is done through `getHearingDateTime()` whereas in the new one it relies on `sscsCaseData.getLatestHearing` which aside from having side-effects is not ordering by `hearingDateTime` but by `id` of the hearing, which cannot be guaranteed will always represent chronological order.

src/main/java/uk/gov/hmcts/reform/sscs/service/PostponementRequestService.java
```diff
@@ -53,4 +58,27 @@ private void ensureSscsDocumentsIsNotNull(SscsCaseData sscsCaseData) {
         final List<SscsDocument> sscsDocuments = (sscsCaseData.getSscsDocument() == null) ? new ArrayList<>() : sscsCaseData.getSscsDocument();
         sscsCaseData.setSscsDocument(sscsDocuments);
     }
+
+    public void addCurrentHearingToExcludeDates(PreSubmitCallbackResponse<SscsCaseData> response) {
+        SscsCaseData caseData = response.getData();
+        Hearing hearing = caseData.getLatestHearing();
```
filippo-possenti-version1, line 64: `getLatestHearings()` has been recently discovered to be a method with side effects. Plus it does additional sorting that may affect performance.

Furthermore, `getLatestHearing()` returns the latest hearing based on its ID, which may not be in chronological order.

- Recommend to check how the removed `setHearingDateToExcludedDate` method was selecting the latest hearing, as it appears to be more stable

src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/actionpostponementrequest/ActionPostponementRequestAboutToSubmitHandler.java
filippo-possenti-version1: Disregard the part where I say that it sorts by `id`. It's counter-intuitive but the `hearingDateTime` is used first and the `id` last. So the only remaining complaint is due to `getLatestHearing` having side-effects, but that should be dealt with in its corresponding project.

This code is fine. Resolving the conversation.

src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/postponementrequest/PostponementRequestAboutToStartHandler.java
filippo-possenti-version1: `getCaseDetails()` is not needed and is fine if we throw a `NullPointerException` in that case.

As for the `isScheduleListingEnabled`, the whole concept of postponement is not released yet and it will be released as part of S&L, so effectively the handler will start being called only when the feature flag will be enabled.
In essence, the whole handler is tied to the S&L feature being enabled so we can work under that assumption.

Resolving this conversation.

src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/postponementrequest/PostponementRequestAboutToStartHandler.java
filippo-possenti-version1: This method has been used in other places and is agreed to be the correct way to determine whether a case is an S&L or GAPS case. Note: GAPS is the legacy system.

As for the feature flag, we rely on the `canHandle` method, which has been now checked and agreed to be correct.

Resolving this conversation

src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/postponementrequest/PostponementRequestAboutToStartHandler.java
filippo-possenti-version1: There are no requirements around the verbiage. We can use the new error message without problems.

Resolving this conversation.

src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/postponementrequest/PostponementRequestAboutToStartHandler.java
filippo-possenti-version1: On line 56 the hearing is actually used so there is purpose in getting a hearing variable. The concerns on `getLatestHearings` side-effects are to be solved in the corresponding library, so this code can remain untouched.

Resolving the conversation.

src/main/java/uk/gov/hmcts/reform/sscs/service/PostponementRequestService.java
filippo-possenti-version1: `getLatestHearings` actually counter-intuitively sorts first by `hearingDateTime` and only afterwards by `id` so the order will be chronological.
As for the side-effects, they are to be handled in the corresponding library. Given that the order is correct, there is therefore no reason to change this code.

Resolving the conversation

filippo-possenti-version1 (dismissed):

amoslv1 (approved):

filippo-possenti-version1 (approved):

dattipoeHMCTS (approved):

PR #3305, opened 2023-06-19
SSCS-11585 - Dropdown List for Language selection values in update to case data 

hajam09:
**Before creating a pull request make sure that:**

Please remove this line and everything above and fill the following sections:

### Change description ###

```
[ ] Yes
[ ] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/caseupdated/CaseUpdatedAboutToSubmitHandler.java
+++ b/src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/caseupdated/CaseUpdatedAboutToSubmitHandler.java
@@ -1,5 +1,6 @@
 package uk.gov.hmcts.reform.sscs.ccd.presubmit.caseupdated;
 
+import static java.util.Objects.isNull;
 import static java.util.Objects.requireNonNull;
 import static org.apache.commons.lang3.StringUtils.isEmpty;
 import static org.apache.commons.lang3.StringUtils.isNotBlank;
@@ -95,6 +96,7 @@ public class CaseUpdatedAboutToSubmitHandler extends ResponseEventsAboutToSubmit
         final boolean hasSuperUserRole = userDetails.hasRole(SUPER_USER);
 
         setCaseCode(preSubmitCallbackResponse, callback, hasSuperUserRole);
+        updateHearingOptionLanguageFromSelectedList(sscsCaseData);
         validateBenefitForCase(preSubmitCallbackResponse, callback, hasSuperUserRole);
         if (!preSubmitCallbackResponse.getErrors().isEmpty()) {
             return preSubmitCallbackResponse;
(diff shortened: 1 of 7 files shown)


amoslv1 (approved):

AndreiTrushkov (dismissed):

src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/caseupdated/CaseUpdatedAboutToSubmitHandler.java
```diff
@@ -323,4 +325,12 @@ private static String resolvePostCode(SscsCaseData sscsCaseData) {
 
         return sscsCaseData.getAppeal().getAppellant().getAddress().getPostcode();
     }
+
+    private void updateHearingOptionLanguageFromSelectedList(SscsCaseData sscsCaseData) {
+        HearingOptions hearingOptions = sscsCaseData.getAppeal().getHearingOptions();
+        if (hearingOptions != null) {
+            DynamicList languagesList = hearingOptions.getLanguagesList();
+            hearingOptions.setLanguages(isNull(languagesList) ? "" : languagesList.getValue().getLabel());
```
DGordonV1, line 333: Is passing an empty string desirable here? What needs to happen for the list to be null here? Is there not a default value we can fall back on instead?

src/test/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/caseupdated/CaseUpdatedAboutToStartHandlerTest.java
```diff
@@ -0,0 +1,99 @@
...
+
+    @Mock
+    private VerbalLanguagesService verbalLanguagesService;
+
+    @InjectMocks
+    private CaseUpdatedAboutToStartHandler handler;
+
+    @Before
+    public void setUp() {
+        sscsCaseData = SscsCaseData.builder().appeal(Appeal.builder().build()).build();
```
DGordonV1, line 54: Could separate the builders out, so Appeal appeal = Appeal.buil...etc then set the .appeal(appeal) for caseData. General rule to keep these separate and not nest them (even when they're simple/empty like this).

DGordonV1 (changes requested):

src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/caseupdated/CaseUpdatedAboutToSubmitHandler.java
hajam09: This is very similar to what's done during the adjournCase event. The original idea was to convert the text field to a drop-down field, but the work required way too many changes. So we decided to keep the original text field (but hide it from the user) and just set it to whatever value was chosen from the dropdown list. If a new case is created in the future, both the text field and dropdown will be empty or null. Besides, the language text field was optional, so setting it to an empty string should be fine.

DGordonV1 (approved):

src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/caseupdated/CaseUpdatedAboutToSubmitHandler.java
DGordonV1: Cool, good explanation, thanks.

ArunabhaChowdhury (approved):

benouaer (approved):

PR #3716, opened 2024-03-28
Update and fix the Cftlib

banderous:
Restores the Cftlib to working order and fixes definition generation and importing.


ruban72 (dismissed):

Andrew-McMahon7 (dismissed):

oleon23 (dismissed):

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "definitions"]
+	path = definitions
+	url = git@github.com:hmcts/sscs-ccd-definitions.git
```
nitinprabhuhmcts, line 3: does this mean we always will checkout master definition ? If we had to deploy local ccd definition changes how do we go about it ?

README.md
```diff
@@ -34,6 +34,32 @@ Run the application by executing:
...
 Extra compose files can be provided in src/cftlib/resources/docker.
 
+#### Cftlib - Docker Desktop v4.30
+
+**_First Time Setup_**
+
+As part of this setup the definitions repository has been added as a git submodule.
+1. `cd definitions`
+2. `git submodule init` & `git submodule update`
+4. `git checkout master` & `git pull`
```
nitinprabhuhmcts, line 44: can we not do this when we run bootWithCcd so we don't have to do this manually ?

README.md
```diff
@@ -34,6 +34,32 @@ Run the application by executing:
 This repo is now integrated with the rse-cft-library. For more information see the bootWithCCD task in build.gradle.
 Extra compose files can be provided in src/cftlib/resources/docker.
 
+#### Cftlib - Docker Desktop v4.30
+
+**_First Time Setup_**
+
+As part of this setup the definitions repository has been added as a git submodule.
```
adusumillipraveen, line 41: Can you move it all into a single script / move to existing script and make it idempotent ?

README.md
```diff
@@ -34,6 +34,32 @@ Run the application by executing:
...
+**On Mac**
+
+1. `sudo nano /etc/hosts`
+2. Add the following line to the file: `127.0.0.1    rse-idam-simulator`
+3. To Save and exit `control + x` to exit, `y` to save, `enter` to confirm
+4. Run `sudo dscacheutil -flushcache; sudo killall -HUP mDNSResponder`
+
+NOTE: Currently local setup fails if `ENABLE_WORK_ALLOCATION` is set to true so change this to false in the build.gradle file.
+
+Run the `bootWithCCD` gradle task. Once you see `Closing the ES REST client`, the application is ready to use.
```
adusumillipraveen, line 59: Don't we still need use line #29  script to run, ay be better to merge them ?

README.md
```diff
@@ -34,6 +34,32 @@ Run the application by executing:
...
+2. `git submodule init` & `git submodule update`
+4. `git checkout master` & `git pull`
+
+You should also run `docker pull hmctspublic.azurecr.io/hmcts/rse/rse-idam-simulator:latest`
+
+_Ensure you have removed all existing / leftover containers_
+
+**On Mac**
+
+1. `sudo nano /etc/hosts`
```
adusumillipraveen, line 52: Same with these, worth moving them to a simple script and make it idempotent. Something like

```
HOST_ENTRY="127.0.0.1    rse-idam-simulator"

if grep -q "$HOST_ENTRY" /etc/hosts; then
  echo "Entry already exists in /etc/hosts"
else
  sudo bash -c "echo '$HOST_ENTRY' >> /etc/hosts"
fi
```

adusumillipraveen (commented):
Minor comments for easy usage

.gitmodules
Andrew-McMahon7: It would clone the entire defs into the folder and then you can change them as normal I believe

src/cftlib/resources/scripts/start.sh
```diff
@@ -15,4 +15,25 @@ loadSecret "LAUNCH_DARKLY_SDK_KEY" "wa-aat" "ld-secret"
...
+echo "Switch to parent folder"
+cd ../
+
+HOST_ENTRY="127.0.0.1    rse-idam-simulator"
+
+if grep -q "$HOST_ENTRY" /etc/hosts; then
+  echo "Entry already exists in /etc/hosts"
+else
+  sudo bash -c "echo '$HOST_ENTRY' >> /etc/hosts"
+fi
```
adusumillipraveen, line 34: Did we miss the dnsflush steps in previous readme?

srochani: Ah good catch, added

adusumillipraveen (dismissed):

README.md
srochani: Added in start.sh

README.md
srochani: Removed

PR #3742, opened 2024-04-16
SSCSSI-194 | Add job chart in prep for setting up cron

adusumillipraveen:
- Use case is to setup a cron job for running one off tasks with any data fixes.
- Currently, other apps use a seperate charts repo and needs environment variables updated everytime there is change in app repo.
-  This setup just adds the job chart as optional with aliasing values from java so that we don't need to duplicate all values over in a separate repo.
- Also, runs the job if u add label `pr-values:job` with a specific `values.job.preview.template.yaml` to configure in preview.

### Checklist

- [x] Does this PR introduce a breaking change


ruban72 (approved):

charts/sscs-tribunals-api/values.yaml
```diff
@@ -139,3 +139,7 @@ ccd-case-document-am-api:
   enabled: false
 ccd-def-cleanup:
   enabled: false
+job:
+  #DO NOT DELETE, Below linn is aliasing all values from Java settings above to job to avoid duplication.
```
nitinprabhuhmcts, line 143: linn ?

nitinprabhuhmcts (approved):

nitinprabhuhmcts (approved):

PR #3793, opened 2024-05-02
Fix preview deployment

markbrooks725:
### Change description ###
Test and fix preview deployment

### DIFF
--- a/Jenkinsfile_CNP
+++ b/Jenkinsfile_CNP
@@ -67,7 +67,8 @@ def secretsForPR = [
                 secret('docmosis-endpoint', 'PDF_SERVICE_CONVERT_URL'),
                 secret('docmosis-api-key', 'PDF_SERVICE_ACCESS_KEY'),
                 secret('AppInsightsInstrumentationKey', 'APPINSIGHTS_INSTRUMENTATIONKEY'),
-                secret('sscs-servicebus-connection-string-tf', 'SERVICE_BUS_CONNECTION_STRING')
+                secret('sscs-servicebus-connection-string-tf', 'SERVICE_BUS_CONNECTION_STRING'),
+                secret('preview-ccd-db-password-2', 'PREVIEW_DB_PASSWORD_2')
         ],
         's2s-${env}'      : [
                 secret('microservicekey-ccd-gw', 'CCD_API_GATEWAY_S2S_SECRET'),
--- a/charts/sscs-tribunals-api/values.ccd.preview.template.yaml
+++ b/charts/sscs-tribunals-api/values.ccd.preview.template.yaml
@@ -741,16 +741,3 @@ servicebus:
         subscriptionNeeded: yes
     queues:
       - name: tribunals-to-hearing-api
-
-wa:
-  enabled: false
-  taskBatchCron:
-    enabled: false
-  reconfigurationBatchCron:
-    enabled: false
-  terminationBatchCron:
-    enabled: false
-aac-manage-case-assignment:
-  enabled: false
-am-org-role-mapping-service:
-  enabled: false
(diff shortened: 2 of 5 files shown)


DGordonV1 (approved):

AndreiTrushkov (dismissed):

Jenkinsfile_CNP
```diff
@@ -67,7 +67,8 @@ def secretsForPR = [
                 secret('docmosis-endpoint', 'PDF_SERVICE_CONVERT_URL'),
                 secret('docmosis-api-key', 'PDF_SERVICE_ACCESS_KEY'),
                 secret('AppInsightsInstrumentationKey', 'APPINSIGHTS_INSTRUMENTATIONKEY'),
-                secret('sscs-servicebus-connection-string-tf', 'SERVICE_BUS_CONNECTION_STRING')
+                secret('sscs-servicebus-connection-string-tf', 'SERVICE_BUS_CONNECTION_STRING'),
+                secret('preview-ccd-db-password-2', 'PREVIEW_DB_PASSWORD_2')
```
adusumillipraveen, line 71: Not a show stopper, but I don't see why we are setting only one API password like this ? where others apps are reading from elsewhere. We only create one db server for all apps in preview, so ideally everything should use same way, a maintenance overhead if not.

charts/sscs-tribunals-api/values.ccd.preview.template.yaml
```diff
@@ -741,3 +741,16 @@ servicebus:
         subscriptionNeeded: yes
     queues:
       - name: tribunals-to-hearing-api
+
+wa:
```
adusumillipraveen, line 745: No point duplicating as most of it is default in https://github.com/hmcts/sscs-tribunals-case-api/blob/9762d8335f8300031e53a8332e0b485005e07998/charts/sscs-tribunals-api/values.yaml#L144-L151

adusumillipraveen (commented):
There is generally a lot of duplication between `values.ccd.preview.template.yaml` and `values.wa.preview.template.yaml`.  wa yaml should only have wa related config and leave ccd yaml for ccd config. Its a maintenance overhead for future changes with this duplication.

sanjayparekh-hmcts (dismissed):

Jenkinsfile_CNP
markbrooks725: For WA we were told to use the Azure PostgreSQL flexible servers, that password is in the vault but this is the only way I could override the default password for the am-role-assignment-service. This can be removed when we update to newer CCD chart.

charts/sscs-tribunals-api/values.ccd.preview.template.yaml
markbrooks725: Yes, I will remove this.

DGordonV1 (approved):

AndreiTrushkov (approved):

Jenkinsfile_CNP
adusumillipraveen: Ahh sure, you don't need them in vault. they are setup as secrets in flux. Application can read like in https://github.com/hmcts/chart-ccd/pull/278/files#diff-5c4fad46c3e248239dcb1170b473d88ff66da225f099d0b6f897d710c1a0fb13R116-R119 . With having it vault, they will be in 2 places and likely go out of sync at some point.

adusumillipraveen: It's encrypted as a secret here : https://github.com/hmcts/cnp-flux-config/blob/master/apps/sscs/preview/sops-secrets/sscs-postgres.enc.yaml

adusumillipraveen: Basically what u removed should work. https://github.com/hmcts/sscs-tribunals-case-api/pull/3793/files#diff-23620b4ffeebecff87d5e2a077b1d0afb80739578e0999577266b0e4cc19b093L335-L339

markbrooks725: @adusumillipraveen the bit I removed worked on the newer CCD chart but didn't work on the older one. It was still using the password from the ccd chart.

adusumillipraveen: This is an additional maintenance though when upgrading, may be leaving a note to remove it on upgrade.

DGordonV1 (approved):

PR #3890, opened 2024-07-12
SSCSSI-352 | Fix prefetching issue with service bus topic message listeners

adusumillipraveen:
### Jira link (if applicable)

https://tools.hmcts.net/jira/browse/SSCSSI-352

### Change description ###

- Currently as the app uses spring jms directly, the default prefetch policy is enabled.
- It prefetches (max 1000) messages into local cache while other listeners are idle
- Message lock duration on subscription is 1 minute. So if all messages in local cache are not processed (which is likely when the cache has more than a few messages) with in 1 minute, service bus will  send this to another listener while this still processes the messages.
- This leads to duplicate message processing and the queues getting piled up with same message for up to 10 times (max redeliveries set on subscription).
- This causes exponential impact and AAT topic goes up to 3000 messages at a given time, making entire AAT blocked. 
- It may turn unmanageable this if it occurs to this level on production.
- A bit more details in Azure documentation  https://learn.microsoft.com/en-us/azure/developer/java/spring-framework/spring-jms-troubleshooting-guide#prefetch-issue 

### Impact on case
Case will be updated with same event over and over again causing data inconsistency .

image-2024-05-24-15-05-14-594

### Volumes

AAT - 11,424 messages processed with duplication in 1 month

Screenshot 2024-07-10 at 09 24 10

Prod - 100 messages processed with duplication in 1 month , less impact due to less volumes of cases.

Screenshot 2024-07-10 at 09 18 39

### Checklist

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/sscs/config/MessagingConfig.java
+++ b/src/main/java/uk/gov/hmcts/reform/sscs/config/MessagingConfig.java
@@ -4,6 +4,7 @@ import javax.jms.ConnectionFactory;
 import javax.net.ssl.SSLContext;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.qpid.jms.JmsConnectionFactory;
+import org.apache.qpid.jms.policy.JmsDefaultPrefetchPolicy;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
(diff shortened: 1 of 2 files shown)


timja (approved):
nice work!

SomeshRA (approved):
LGTM

nitinprabhuhmcts (approved):

PR #3953, opened 2024-09-10
SSCSCI-1028: move mrd reference lookup logic to common

benouaer:
### Jira link
- See [SSCSCI-1028](https://tools.hmcts.net/jira/browse/SSCSCI-1028)

### Change description
- move mrdReference/dialect lookup logic to common for use by both tribunals and hearings APIs


nitinprabhuhmcts (approved):

src/main/java/uk/gov/hmcts/reform/sscs/util/DynamicListLanguageUtil.java
```diff
@@ -45,19 +44,6 @@ private List<DynamicListItem> generateInterpreterLanguage() {
 
     @NotNull
     public DynamicListItem getLanguageDynamicListItem(Language language) {
```
nitinprabhuhmcts, line 46: Do we need a test I know it's in common but from util point we need to have a test ?

KatKovacs1 (dismissed):

src/main/java/uk/gov/hmcts/reform/sscs/util/DynamicListLanguageUtil.java
benouaer: There is already a test. I've cleaned it up a bit, so it should appear in the PR now

nitinprabhuhmcts (dismissed):

hashimalisolirius (approved):

PR #3960, opened 2024-09-16
SSCSCI-1036

tomxelliott:
### Jira link
https://tools.hmcts.net/jira/browse/SSCSCI-1036

### Change description
- Move fetching of env var secrets from script into gradle file
- Keep old /start.sh file for backward compatibility
- Update README

Screenshot 2024-09-16 at 15 34 49


benouaer (approved):

nilay913 (approved):

nitinprabhuhmcts (approved):

linusnorton:
Previously I think we just put the secrets in .aat-env rather than all the env vars configuring URLs and whatnot but I guess it makes sense. We will need to run this passed the other devs first I think.

linusnorton (approved):

src/cftlib/resources/scripts/start.sh
```diff
@@ -1,19 +1,3 @@
 #!/bin/bash
```
sanjayparekh-hmcts, line 2: maybe we just need a small comment here to indicate to the dev that this scripting has been deprecated and that they should view the README

sanjayparekh-hmcts (approved):

PR #4172, opened 2024-11-25
add letter testing setup

benouaer:
### Jira link

See [PROJ-XXXXXX](https://tools.hmcts.net/jira/browse/PROJ-XXXXXX)

### Checklist


Jenkinsfile_CNP
```diff
@@ -178,13 +179,27 @@ withPipeline(type, product, component) {
     }
 
     afterSuccess('smoketest:preview') {
+        if (notificationTestsEnabled) {
+            runNotificationTests()
+        }
```
linusnorton, line 184: maybe we want an else statement here to cut the run time?

build.gradle
```diff
@@ -118,9 +118,19 @@ tasks.register('functional', Test) {
     maxParallelForks = Runtime.runtime.availableProcessors()
 }
 
+tasks.register('notifications', Test) {
+    group = 'Notification Tests'
```
linusnorton, line 122: verification?

Jenkinsfile_CNP
benouaer: what would we do in the else branch?
currently, the tests only run if you add the run-notification-tests label, and you also have the option to only generate one or a few notifications rather than the whole set

PR #4323, opened 2025-01-22
SSCSCI-1588

nilay913:
### Jira link

See [SSCSCI-1588](https://tools.hmcts.net/jira/browse/SSCSCI-1588)

### Change description
Pointing to new templates to replace hardcoded "NI No" with <<label>> placeholder for draft and final decision notice

### Checklist


src/IntegrationTests/java/uk/gov/hmcts/reform/sscs/config/DocumentConfigurationTest.java
```diff
@@ -39,13 +39,13 @@ public class DocumentConfigurationTest {
     public static Object[][] documentParameters() {
         return new Object[][] {
                 {LanguagePreference.ENGLISH, EventType.CORRECTION_GRANTED, Optional.empty(), "TB-SCS-GNO-ENG-corrected-decision-notice.docx"},
```
benouaer, line 41: should we get these from the yaml as well?

src/main/resources/config/application.properties.old
```diff
@@ -71,13 +71,13 @@ docmosis.uri=${DOCMOSIS_SERVICE_BASE_URL:https://docmosis.aat.platform.hmcts.net
 documents.english.CORRECTION_GRANTED=TB-SCS-GNO-ENG-corrected-decision-notice.docx
```
benouaer, line 71: is this file still being used?

benouaer (approved):

benouaer:
general question not necessarily about your changes:
- do we really need all those yaml files? surely we can use the main one, and only create a separate one when we need to override some properties
- do we need to specify the name of the template in all those places? should we not just get it from the yaml?

nilay913:
> general question not necessarily about your changes:
> 
> * do we really need all those yaml files? surely we can use the main one, and only create a separate one when we need to override some properties
> * do we need to specify the name of the template in all those places? should we not just get it from the yaml?

I don't think they're really needed, maybe as part of a clean up ticket we can remove some seemingly redundant ones? i.e. application.properties.old

src/main/resources/config/application.properties.old
nilay913: unsure, updated it just in case

src/IntegrationTests/java/uk/gov/hmcts/reform/sscs/config/DocumentConfigurationTest.java
nilay913: maybe a task as part of clean up too?

PR #5244, opened 2026-04-28
SSCSCI-2490 Remove HEF notification for CM/UC cases

hellobasitsiddiqui:
## Summary
- Suppresses the HEF notification (`TB-SCS-GNO-ENG-00784-v2.docx`) from being automatically sent on `UPDATE_OTHER_PARTY_DATA` events for CM and UC cases
- Gated behind the existing `feature.cm-other-party-confidentiality.enabled` flag — flag off = no change, flag on = HEF suppressed for CM/UC benefit types

## Test plan

- [ ] Existing unit tests passing


nitinprabhuhmcts (approved):

src/main/java/uk/gov/hmcts/reform/sscs/tyanotifications/factory/CcdNotificationWrapper.java
```diff
@@ -282,10 +290,14 @@ private boolean isNotificationEventValidToSendToJointParty(SscsCaseData newSscsC
     }
 
     private boolean isNotificationEventValidToSendToOtherPartySubscription(Subscription subscription, boolean isSendNewOtherPartyNotification, SscsCaseData newSscsCaseData, NotificationEventType notificationEventType, String partyMember) {
+        boolean isSuppressedForConfidentiality = cmOtherPartyConfidentialityEnabled
```
stuart-hmcts: Could inline either as is or as a method call. We only want to do this for UPDATE_OTHER_PARTY_DATA event scenario.

KatKovacs1 (approved):

src/main/java/uk/gov/hmcts/reform/sscs/tyanotifications/factory/CcdNotificationWrapper.java
```diff
@@ -282,10 +290,14 @@ private boolean isNotificationEventValidToSendToJointParty(SscsCaseData newSscsC
...
+            && newSscsCaseData.getAppeal() != null
+            && OtherPartyDataUtil.isValidBenefitTypeForConfidentiality(newSscsCaseData.getAppeal().getBenefitType());
         boolean isValid = isValidSubscriptionOrIsMandatoryLetter(subscription, responseWrapper.getNotificationEventType())
             && (EVENTS_VALID_FOR_ALL_ENTITIES.contains(notificationEventType)
             || EVENTS_VALID_FOR_OTHER_PARTY.contains(notificationEventType)
-            || (UPDATE_OTHER_PARTY_DATA.equals(notificationEventType) && isSendNewOtherPartyNotification));
+            || (UPDATE_OTHER_PARTY_DATA.equals(notificationEventType) && isSendNewOtherPartyNotification
```
stuart-hmcts: I'm not sure on the design of notifications as there is a lot going on. However this specific method seems generic based on enums. Adding bespoke logic like this probably isn't a good pattern to follow in terms of future maintainability or growth. Within NotificationService there is an isEventAllowedToProceedWithValidData method which suppresses notification based on specific criteria. It's not a particularly nice method but it does seem to be where specific exceptions are added. Although I can't say for certain when to add logic to which.

@nitinprabhuhmcts - can you advise? I'm just conscious that I've added similar code in the NotificationService and Basit has added code here and it would be better to be consistent.

src/test/java/uk/gov/hmcts/reform/sscs/tyanotifications/factory/CcdNotificationWrapperTest.java
```diff
@@ -602,6 +613,27 @@ public void givenUpdateOtherPartyDataEventAndSendNotificationFlagIsSetInOtherPar
             .containsOnly("2", "3");
     }
 
+    @Test
+    public void givenUpdateOtherPartyDataEventAndCmFlagOnAndChildSupportCase_thenReturnEmptySubscription() {
+        ccdNotificationWrapper = buildNotificationWrapperWithOtherPartyBenefitTypeAndFlag(UPDATE_OTHER_PARTY_DATA, buildOtherPartyData(true, false, false), "childSupport", true);
+        List<SubscriptionWithType> subsWithTypeList = ccdNotificationWrapper.getOtherPartySubscriptions(ccdNotificationWrapper.getNewSscsCaseData(), ccdNotificationWrapper.getNotificationType());
+        Assertions.assertThat(subsWithTypeList).isEmpty();
```
nitinprabhuhmcts: can be inlined `assertThat(subsWithTypeList).isEmpty();`

src/main/java/uk/gov/hmcts/reform/sscs/tyanotifications/factory/CcdNotificationWrapper.java
nitinprabhuhmcts: I agree better to be consistent and move the logic there `isEventAllowedToProceedWithValidData`

src/main/java/uk/gov/hmcts/reform/sscs/tyanotifications/factory/CcdNotificationWrapper.java
hellobasitsiddiqui: Done — extracted to a private method `isHefSuppressedForConfidentiality`.

src/test/java/uk/gov/hmcts/reform/sscs/tyanotifications/factory/CcdNotificationWrapperTest.java
hellobasitsiddiqui: Done — switched to static import.

src/main/java/uk/gov/hmcts/reform/sscs/tyanotifications/factory/CcdNotificationWrapper.java
hellobasitsiddiqui: the `isEventAllowedToProceedWithValidData` suggestion — the issue is that method kills the entire notification event, so `appellant` and `rep` would also stop getting their notifications. we only want to suppress the HEF for `other parties`, so keeping it in `isNotificationEventValidToSendToOtherPartySubscription` gives us that granularity. @stuart-hmcts @nitinprabhuhmcts

hellobasitsiddiqui: addressed

src/main/java/uk/gov/hmcts/reform/sscs/tyanotifications/service/NotificationService.java
```diff
@@ -297,6 +302,14 @@ boolean isNotificationStillValidToSendSetAsideRequest(SscsCaseData caseData, Not
 
     private boolean isEventAllowedToProceedWithValidData(NotificationWrapper notificationWrapper,
                                                          NotificationEventType notificationType) {
+        if (UPDATE_OTHER_PARTY_DATA.equals(notificationType)
+            && cmOtherPartyConfidentialityEnabled
+            && notificationWrapper.getNewSscsCaseData().getAppeal() != null
+            && OtherPartyDataUtil.isValidBenefitTypeForConfidentiality(notificationWrapper.getNewSscsCaseData().getAppeal().getBenefitType())) {
```
stuart-hmcts: I think we are just interested in Child Benefit and UC. This checks against SSCS2 and SCS5 which covers a lot more benefit types.

hellobasitsiddiqui: confirmed with Mindip (BA) — it should be CM and UC only. updated the logic to use a private `isCmOrUcCase` method that checks specifically for `childSupport` and `UC` benefit codes rather than the broader `isValidBenefitTypeForConfidentiality` which covers all types.

stuart-hmcts: There is already an "isBenefitType" helper method on SscsCaseData. eg.

notificationWrapper.getNewSscsCaseData().isBenefitType(CHILD_SUPPORT)

If you use that then you don't need to check if the appeal is null (it does that) and it will make the code more succinct.

src/main/java/uk/gov/hmcts/reform/sscs/tyanotifications/service/NotificationService.java
```diff
@@ -297,6 +301,14 @@ boolean isNotificationStillValidToSendSetAsideRequest(SscsCaseData caseData, Not
 
     private boolean isEventAllowedToProceedWithValidData(NotificationWrapper notificationWrapper,
                                                          NotificationEventType notificationType) {
+        if (UPDATE_OTHER_PARTY_DATA.equals(notificationType)
+            && cmOtherPartyConfidentialityEnabled
+            && notificationWrapper.getNewSscsCaseData().getAppeal() != null
+            && isCmOrUcCase(notificationWrapper.getNewSscsCaseData().getAppeal().getBenefitType())) {
+            log.info("Suppressing HEF notification for CM/UC case id {}.", notificationWrapper.getCaseId());
```
stuart-hmcts: At this stage we know what the benefit type is so rather than having CM/UC you could inject the actual benefit type. That would make the log statement more relative.

stuart-hmcts (approved):

src/main/java/uk/gov/hmcts/reform/sscs/tyanotifications/service/NotificationService.java
```diff
@@ -422,6 +434,12 @@ private boolean hasHearingChannelNotChanged(HearingDetails newHearing, HearingDe
         return oldHearingChannel.equals(newHearingChannel);
     }
 
+    private boolean isCmOrUcCase(BenefitType benefitType) {
+        return benefitType != null
+            && (Benefit.CHILD_SUPPORT.getShortName().equals(benefitType.getCode())
+                || Benefit.UC.getShortName().equals(benefitType.getCode()));
+    }
```
KatKovacs1: Wouldn't something like this work?
```
private boolean isCmOrUcCase(SscsCaseData caseData) {
        return caseData.isBenefitType(Benefit.CHILD_SUPPORT) || caseData.isBenefitType(Benefit.UC);
    }
```

src/main/java/uk/gov/hmcts/reform/sscs/tyanotifications/service/NotificationService.java
hellobasitsiddiqui: done - switched to `isBenefitType(CHILD_SUPPORT)` and `isBenefitType(UC)` directly in the condition. also removed the `appeal != null` since `isBenefitType` handles nulls internally.

src/main/java/uk/gov/hmcts/reform/sscs/tyanotifications/service/NotificationService.java
hellobasitsiddiqui: done - now logs the actual benefit type e.g. `childSupport` or `UC` instead of the hardcoded string.

src/main/java/uk/gov/hmcts/reform/sscs/tyanotifications/service/NotificationService.java
hellobasitsiddiqui: done - went one step further and inlined it directly in the condition rather than keeping a separate helper so no `isCmOrUcCase` method needed at all.

KatKovacs1 (approved):

stuart-hmcts (approved):

PR #5334, opened 2026-06-18
SSCSCI-2659 Add other party recipient selection to Issue Direction Notice

hellobasitsiddiqui:
### Jira link

[SSCSCI-2659](https://tools.hmcts.net/jira/browse/SSCSCI-2659)

### Change description

Lets the judge/LO pick which other party receives the Issue Direction Notice on CM/UC cases, instead of it going to all of them where more than one exists. Same recipient-selection approach as the Issue Generic Letter / HEF events, and selecting the same party twice is blocked.

Draft for now - the CCD show-condition that puts the picker on the screen is parked pending a business decision on scope (confidentiality directions only vs any CM/UC direction), so that field row isn't in yet.

### Testing done

Unit tests across the notification filter (only the selected other parties get a notification record, EN + CY), the about-to-start picker seeding, and the mid-event duplicate guard. Local Playwright UI check to follow once the screen-config row lands.

### Checklist

- [x] commit messages are meaningful and follow good commit message guidelines
- [x] tests have been updated / new tests has been added (if needed)


definitions/benefit/sheets/AuthorisationCaseField/AuthorisationCaseField.json
```diff
@@ -1503,6 +1503,9 @@
   {"LiveFrom": "01/01/2018", "CaseTypeID": "Benefit", "CaseFieldID": "otherPartySelection", "AccessProfile": "caseworker-sscs-clerk", "CRUD": "CRUD"},
   {"LiveFrom": "01/01/2018", "CaseTypeID": "Benefit", "CaseFieldID": "otherPartySelection", "AccessProfile": "caseworker-sscs-superuser", "CRUD": "CRUD"},
   {"LiveFrom": "01/01/2018", "CaseTypeID": "Benefit", "CaseFieldID": "otherPartySelection", "AccessProfile": "caseworker-sscs-systemupdate", "CRUD": "CRUD"},
+  {"LiveFrom": "01/01/2018", "CaseTypeID": "Benefit", "CaseFieldID": "otherPartySelection", "AccessProfile": "caseworker-sscs-judge", "CRUD": "CRUD"},
```
stuart-hmcts: Please correct the dates to be current.

definitions/benefit/sheets/CaseEventToFields/CaseEventToFields.json
```diff
@@ -2275,14 +2275,28 @@
...
     "PageLabel": "Issue direction",
     "RetainHiddenValue": "Yes"
   },
+  {
+    "CaseEventID": "directionIssued",
+    "CaseFieldID": "otherPartySelection",
+    "CaseTypeID": "Benefit",
+    "DisplayContext": "MANDATORY",
+    "FieldShowCondition": "sendDirectionNoticeToOtherParty = \"Yes\"",
+    "LiveFrom": "01/01/2018",
```
hashimalisolirius: Could we update this date?

definitions/benefit/sheets/CaseEventToFields/CaseEventToFields.json
```diff
@@ -2275,14 +2275,28 @@
     "CaseFieldID": "sendDirectionNoticeToOtherParty",
     "CaseTypeID": "Benefit",
     "DisplayContext": "MANDATORY",
-    "FieldShowCondition": "confidentialityType=\"confidential\" AND hasOtherParties = \"Yes\"",
+    "FieldShowCondition": "hasOtherParties = \"Yes\"",
     "LiveFrom": "01/01/2018",
```
stuart-hmcts, line 2291: Please correct the date to be current.

src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/directionissued/DirectionIssuedAboutToStartHandler.java
```diff
@@ -150,6 +152,22 @@ private void clearFields(SscsCaseData sscsCaseData) {
         sscsCaseData.setSendDirectionNoticeToAppellantOrAppointee(null);
     }
 
+    // SSCSCI-2659: build the other party picker on about-to-start only - a mid-event would wipe the selection.
+    // CM/UC with other parties, otherwise clear it. Same as issue generic letter.
+    private void setOtherPartySelectionList(SscsCaseData sscsCaseData) {
+        if (cmDirectionTypesConfidentiality
+            && isBenefitTypeChildSupportOrUc(sscsCaseData)
+            && OtherPartyDataUtil.isOtherPartyPresent(sscsCaseData)) {
```
stuart-hmcts: Consider a static import on OtherPartyDataUtil

src/test/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/directionissued/DirectionIssuedAboutToStartHandlerTest.java
```diff
@@ -491,6 +491,37 @@ void givenAValidCallbackType_thenVerifyAllPartiesOnTheCase() {
...
+    void givenNonCmCase_thenOtherPartySelectionNotSeeded() {
+        handler = new DirectionIssuedAboutToStartHandler(false, true, idamService);
+        sscsCaseData.getAppeal().setBenefitType(BenefitType.builder().code(PIP.getShortName()).build());
+        CcdValue<OtherParty> op = CcdValue.<OtherParty>builder()
+                .value(OtherParty.builder().id("1").name(Name.builder().firstName("Harry").lastName("Kane").build()).build()).build();
+        sscsCaseData.setOtherParties(List.of(op));
+
+        handler.handle(ABOUT_TO_START, callback, USER_AUTHORISATION);
+
+        assertNull(sscsCaseData.getOtherPartySelection());
```
hashimalisolirius: Think it's better for this to be assertThat().isNull()

src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/directionissued/DirectionIssuedMidEventHandler.java
```diff
@@ -47,6 +48,12 @@ public PreSubmitCallbackResponse<SscsCaseData> handle(CallbackType callbackType,
         SscsCaseData caseData = callback.getCaseDetails().getCaseData();
         PreSubmitCallbackResponse<SscsCaseData> errorResponse = new PreSubmitCallbackResponse<>(caseData);
 
+        // SSCSCI-2659: stop the caseworker picking the same other party more than once on the recipient list.
```
stuart-hmcts: My view is that we don't need comments - the code should be readable enough as is.

hashimalisolirius (approved):

src/main/java/uk/gov/hmcts/reform/sscs/tyanotifications/factory/CcdNotificationWrapper.java
```diff
@@ -172,6 +176,37 @@ && isNotificationEventValidToSendToOtherPartySubscription(otherParty.getOtherPar
         return otherPartySubscription;
     }
 
+    // SSCSCI-2659: on a direction notice, only notify the other parties the caseworker actually picked.
```
stuart-hmcts: I don' think we need all the comments. The code should be readable as is.

src/main/java/uk/gov/hmcts/reform/sscs/tyanotifications/factory/CcdNotificationWrapper.java
```diff
@@ -172,6 +176,37 @@ && isNotificationEventValidToSendToOtherPartySubscription(otherParty.getOtherPar
...
+            .map(DynamicListItem::getCode)
+            .filter(Objects::nonNull)
+            .collect(Collectors.toList());
+        // No specific other party chosen (picker not shown, or seeded row with nothing selected) -> don't filter.
+        if (selectedCodes.isEmpty()) {
+            return true;
+        }
+        // Match on any of the party's ids: the picker and the appointee/other-party branch logic can disagree
+        // on which id to use when appointee data is malformed, so accepting either avoids dropping a valid pick.
+        for (String partyId : partyIds) {
```
stuart-hmcts: partyIds is a param into this method. We don't control what is sent. Should be a null check before the for loop in order to avoid a potential NPE

src/main/java/uk/gov/hmcts/reform/sscs/tyanotifications/factory/CcdNotificationWrapper.java
```diff
@@ -172,6 +176,37 @@ && isNotificationEventValidToSendToOtherPartySubscription(otherParty.getOtherPar
...
+            if (Objects.nonNull(partyId)
```
stuart-hmcts: Static import on Objects

src/main/java/uk/gov/hmcts/reform/sscs/tyanotifications/factory/CcdNotificationWrapper.java
```diff
@@ -172,6 +176,37 @@ && isNotificationEventValidToSendToOtherPartySubscription(otherParty.getOtherPar
...
+        }
+        List<String> selectedCodes = emptyIfNull(caseData.getOtherPartySelection()).stream()
+            .map(CcdValue::getValue)
+            .map(OtherPartySelectionDetails::getOtherPartiesList)
+            .filter(Objects::nonNull)
+            .map(DynamicList::getValue)
+            .filter(Objects::nonNull)
+            .map(DynamicListItem::getCode)
+            .filter(Objects::nonNull)
+            .collect(Collectors.toList());
```
stuart-hmcts: Can simplify to toList

definitions/benefit/sheets/CaseEventToFields/CaseEventToFields.json
```diff
@@ -2275,14 +2275,28 @@
     "CaseFieldID": "sendDirectionNoticeToOtherParty",
     "CaseTypeID": "Benefit",
     "DisplayContext": "MANDATORY",
```
stuart-hmcts, line 2289: When I don't enter a value the error message is "Field is required". This should be more descriptive.

src/test/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/directionissued/DirectionIssuedAboutToStartHandlerTest.java
hellobasitsiddiqui: done

src/main/java/uk/gov/hmcts/reform/sscs/tyanotifications/factory/CcdNotificationWrapper.java
```diff
@@ -152,17 +152,21 @@ private List<SubscriptionWithType> filterOtherPartySubscription(SscsCaseData new
 
         boolean isSendNewOtherPartyNotification = YesNo.isYes(otherParty.getSendNewOtherPartyNotification());
 
+        String appointeeId = otherParty.getAppointee() != null ? otherParty.getAppointee().getId() : null;
         if (hasAppointee(otherParty.getAppointee(), otherParty.getIsAppointee())
-            && isNotificationEventValidToSendToOtherPartySubscription(otherParty.getOtherPartyAppointeeSubscription(), isSendNewOtherPartyNotification, newSscsCaseData, notificationEventType, ConfidentialityPartyMembers.OTHER_PARTY_APPOINTEE.getCode())) {
+            && isNotificationEventValidToSendToOtherPartySubscription(otherParty.getOtherPartyAppointeeSubscription(), isSendNewOtherPartyNotification, newSscsCaseData, notificationEventType, ConfidentialityPartyMembers.OTHER_PARTY_APPOINTEE.getCode())
+            && isOtherPartySelectedForDirectionNotice(newSscsCaseData, notificationEventType, otherParty.getAppointee().getId(), otherParty.getId())) {
```
stuart-hmcts, line 159: This seems like a very scenario specific bit of logic for a generic bit of code. If the intent is to only send to selected other parties for issue directions notice then it might be better to look at applying the logic in the notification service code only - there are plenty of areas in that where specific decisions are made re. what to send and what not to send.

src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/directionissued/DirectionIssuedAboutToStartHandler.java
```diff
@@ -150,6 +152,22 @@ private void clearFields(SscsCaseData sscsCaseData) {
         sscsCaseData.setSendDirectionNoticeToAppellantOrAppointee(null);
     }
 
+    // SSCSCI-2659: build the other party picker on about-to-start only - a mid-event would wipe the selection.
+    // CM/UC with other parties, otherwise clear it. Same as issue generic letter.
+    private void setOtherPartySelectionList(SscsCaseData sscsCaseData) {
+        if (cmDirectionTypesConfidentiality
```
stuart-hmcts, line 158: We only populate the other party selection under a very specific set of circumstances - FF is on and it's CS or UC etc. However, from what I can tell, the other party selection doesn't have these restrictions. So what happens if it's another appeal type? Does the drop down show, and if so how is it populated? Or should the other party select have similar restrictions?

src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/directionissued/DirectionIssuedAboutToStartHandler.java
hellobasitsiddiqui: done

src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/directionissued/DirectionIssuedMidEventHandler.java
hellobasitsiddiqui: done

src/main/java/uk/gov/hmcts/reform/sscs/tyanotifications/factory/CcdNotificationWrapper.java
hellobasitsiddiqui: done

src/main/java/uk/gov/hmcts/reform/sscs/tyanotifications/factory/CcdNotificationWrapper.java
hellobasitsiddiqui: done

src/main/java/uk/gov/hmcts/reform/sscs/tyanotifications/factory/CcdNotificationWrapper.java
hellobasitsiddiqui: done

src/main/java/uk/gov/hmcts/reform/sscs/tyanotifications/factory/CcdNotificationWrapper.java
hellobasitsiddiqui: done

definitions/benefit/sheets/AuthorisationCaseField/AuthorisationCaseField.json
hellobasitsiddiqui: done

definitions/benefit/sheets/CaseEventToFields/CaseEventToFields.json
hellobasitsiddiqui: done - this one was already updated to the current date in an earlier commit

hellobasitsiddiqui: correction on my last reply - this date is on the sendDirectionNoticeToOtherParty row, which was still 01/01/2018. now bumped to 23/06/2026 in 19d47b2f7

definitions/benefit/sheets/CaseEventToFields/CaseEventToFields.json
hellobasitsiddiqui: done - made the field OPTIONAL and added a descriptive mid-event validation ("Select whether to send the direction notice to the other party") instead of the generic "Field is required"

definitions/benefit/sheets/CaseEventToFields/CaseEventToFields.json
hellobasitsiddiqui: done

src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/directionissued/DirectionIssuedAboutToStartHandler.java
hellobasitsiddiqui: the picker only populates for cs/uc but its visibility wasn't restricted the same way, so another benefit type with other parties could show an empty dropdown.

couldn't tie the show condition to benefit type here (can't reference appeal.benefitType.code on this event - fails the import), so i set an "allowed" flag in the about-to-start handler (true only for cs/uc with other parties) and pointed the radio's show condition at it. no empty dropdown now.

src/main/java/uk/gov/hmcts/reform/sscs/tyanotifications/factory/CcdNotificationWrapper.java
hellobasitsiddiqui: i did look at moving it to the notification service but think it's at the correct place - the wrapper is where we build the subscription list and if an other party isn't selected we just never create a subscription for them. that's what guarantees no correspondence record gets written for them. moving it to the service would mean creating the subscription then suppressing the send which still leaves a record.
it also sits next to the confidentiality check the same branches already use so the "**should this other party get this notice**" logic stays in one place. the isResendTo path looks similar but only fires on reissue (resend another copy) ours is the initial send - so not quite the same.

src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/directionissued/DirectionIssuedAboutToStartHandler.java
stuart-hmcts: You should be able to register appeal/benefittype against the event to use it - that's what we do elsewhere. It's the standard practice and saves having to introduce another field.

hellobasitsiddiqui: good point - done. registered appeal against the event (like subscriptionUpdated does) and put the benefit check back on the show condition, dropped the extra field. pipeline's green incl the preview build so it imports fine.

definitions/benefit/sheets/CaseEventToFields/CaseEventToFields.json
```diff
@@ -2274,15 +2286,29 @@
     "CaseEventID": "directionIssued",
     "CaseFieldID": "sendDirectionNoticeToOtherParty",
     "CaseTypeID": "Benefit",
-    "DisplayContext": "MANDATORY",
-    "FieldShowCondition": "confidentialityType=\"confidential\" AND hasOtherParties = \"Yes\"",
-    "LiveFrom": "01/01/2018",
+    "DisplayContext": "OPTIONAL",
+    "FieldShowCondition": "hasOtherParties = \"Yes\" AND (appeal.benefitType.code=\"childSupport\" OR appeal.benefitType.code=\"UC\")",
```
stuart-hmcts: Why did we need to change this? Doesn't this option still apply to the other benefit types? And doesn't the confidentiality check still apply?

src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/directionissued/DirectionIssuedMidEventHandler.java
```diff
@@ -47,6 +48,16 @@ public PreSubmitCallbackResponse<SscsCaseData> handle(CallbackType callbackType,
         SscsCaseData caseData = callback.getCaseDetails().getCaseData();
         PreSubmitCallbackResponse<SscsCaseData> errorResponse = new PreSubmitCallbackResponse<>(caseData);
 
+        if (isYes(caseData.getHasOtherParties()) && caseData.getSendDirectionNoticeToOtherParty() == null) {
```
stuart-hmcts: I think that this check only applies to UC/CS. "send direction notice to other party" is also initialised to null in the "about to start" handler which also processes mid event requests; so introduces a race condition.

src/main/java/uk/gov/hmcts/reform/sscs/tyanotifications/factory/CcdNotificationWrapper.java
```diff
@@ -172,6 +177,35 @@ && isNotificationEventValidToSendToOtherPartySubscription(otherParty.getOtherPar
...
+        List<String> selectedCodes = emptyIfNull(caseData.getOtherPartySelection()).stream()
+            .map(CcdValue::getValue)
+            .map(OtherPartySelectionDetails::getOtherPartiesList)
+            .filter(Objects::nonNull)
+            .map(DynamicList::getValue)
+            .filter(Objects::nonNull)
+            .map(DynamicListItem::getCode)
+            .filter(Objects::nonNull)
+            .toList();
+        if (selectedCodes.isEmpty()) {
```
stuart-hmcts: If we select No for issue to other parties, is there are risk that this check will result in the notification being send to all other parties? ie. there will be no "other party selection" because they selected No, but this says that if there are no selected codes then return true - so presumably all other parties will receive the notification which is the opposite of what was asked?

src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/directionissued/DirectionIssuedMidEventHandler.java
ReplyJordan: Fair enough, added condition to check for UC/CS

definitions/benefit/sheets/CaseEventToFields/CaseEventToFields.json
ReplyJordan: Fair enough, I've added confidentiality check back

src/main/java/uk/gov/hmcts/reform/sscs/tyanotifications/factory/CcdNotificationWrapper.java
hashimalisolirius: Need to double check this logic with POs, it shouldn't go into this logic if no is selected for other party (getEligiblePartyMembersInTheCaseToSendNotification should return false when no is selected). But if user selects yes for other parties, it will send to all other parties if no other party is selected in the dropdown.

definitions/benefit/sheets/AuthorisationCaseField/AuthorisationCaseField.json
```diff
@@ -1503,6 +1503,9 @@
   {"LiveFrom": "01/01/2018", "CaseTypeID": "Benefit", "CaseFieldID": "otherPartySelection", "AccessProfile": "caseworker-sscs-clerk", "CRUD": "CRUD"},
   {"LiveFrom": "01/01/2018", "CaseTypeID": "Benefit", "CaseFieldID": "otherPartySelection", "AccessProfile": "caseworker-sscs-superuser", "CRUD": "CRUD"},
   {"LiveFrom": "01/01/2018", "CaseTypeID": "Benefit", "CaseFieldID": "otherPartySelection", "AccessProfile": "caseworker-sscs-systemupdate", "CRUD": "CRUD"},
+  {"LiveFrom": "23/06/2026", "CaseTypeID": "Benefit", "CaseFieldID": "otherPartySelection", "AccessProfile": "caseworker-sscs-judge", "CRUD": "CRUD"},
+  {"LiveFrom": "23/06/2026", "CaseTypeID": "Benefit", "CaseFieldID": "otherPartySelection", "AccessProfile": "caseworker-sscs-registrar", "CRUD": "CRUD"},
+  {"LiveFrom": "23/06/2026", "CaseTypeID": "Benefit", "CaseFieldID": "otherPartySelection", "AccessProfile": "caseworker-sscs-teamleader", "CRUD": "CRUD"},
```
nitinprabhuhmcts: `caseworker-sscs-teamleader` are you sure about this as this looks like work allocation role ?

src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/directionissued/DirectionIssuedAboutToStartHandler.java
```diff
@@ -150,9 +152,23 @@ private void clearFields(SscsCaseData sscsCaseData) {
         sscsCaseData.setSendDirectionNoticeToAppellantOrAppointee(null);
     }
 
+    private void setOtherPartySelectionList(SscsCaseData sscsCaseData) {
+        if (cmDirectionTypesConfidentiality
```
nitinprabhuhmcts, line 158: some logging would be helpful when debugging in prod

src/main/java/uk/gov/hmcts/reform/sscs/tyanotifications/factory/CcdNotificationWrapper.java
```diff
@@ -172,6 +178,37 @@ && isNotificationEventValidToSendToOtherPartySubscription(otherParty.getOtherPar
         return otherPartySubscription;
     }
 
+    private boolean isOtherPartySelectedForDirectionNotice(SscsCaseData caseData, NotificationEventType eventType, String... partyIds) {
```
nitinprabhuhmcts: can we simplify this method ?
extract the code construction/check into a small helper. That makes the business rule much easier to read

nitinprabhuhmcts (approved):

src/main/java/uk/gov/hmcts/reform/sscs/ccd/presubmit/directionissued/DirectionIssuedAboutToStartHandler.java
```diff
@@ -150,9 +154,24 @@ private void clearFields(SscsCaseData sscsCaseData) {
...
+            && isBenefitTypeChildSupportOrUc(sscsCaseData)
+            && isOtherPartyPresent(sscsCaseData)) {
+            List<DynamicListItem> listOptions = new ArrayList<>();
+            addOtherPartiesToListOptions(sscsCaseData, listOptions, true);
+            List<CcdValue<OtherPartySelectionDetails>> selection = new ArrayList<>();
+            selection.add(new CcdValue<>(new OtherPartySelectionDetails(new DynamicList(null, listOptions))));
+            sscsCaseData.setOtherPartySelection(selection);
+        } else {
+            log.warn("Other party selection not set for case {}", sscsCaseData.getCcdCaseId());
```
stuart-hmcts: Isn't this the happy path route for the majority of appeals? Not sure we need a log here as it will spam the logs for no added benefit. If we do want a log statement then I'm not sure it's a warn.

src/main/java/uk/gov/hmcts/reform/sscs/tyanotifications/factory/CcdNotificationWrapper.java
```diff
@@ -1,13 +1,15 @@
 package uk.gov.hmcts.reform.sscs.tyanotifications.factory;
 
 import static java.util.Objects.isNull;
+import static java.util.Objects.nonNull;
```
stuart-hmcts: Don't think this import is used? Surprised that got past checkstyle.

KatKovacs1 (approved):

src/main/java/uk/gov/hmcts/reform/sscs/tyanotifications/factory/CcdNotificationWrapper.java
KatKovacs1: It's used in the filtering
`.filter(Objects::nonNull)`

stuart-hmcts: it's greyed out in my IDE suggesting it's not needed and if I remove it the code still compiles ok.

I think "import java.util.*;" covers it?

stuart-hmcts (approved):

nitinprabhuhmcts (approved):

nitinprabhuhmcts (approved):

stuart-hmcts (approved):

PR #5352, opened 2026-06-26
Use az identity only when running on jenkins pipeline and not on local

nitinprabhuhmcts:
> The MSI endpoint 169.254.169.254 is only accessible from within Azure. If you are running ` az login --identity`  command from outside Azure, it won’t be able to reach this endpoint. if you are running the CLI from a local environment or outside Azure, you won't be able to use the --identity parameter.


bin/create-xlsx.sh
```diff
@@ -3,14 +3,15 @@
...
+USE_IDENTITY=${4:-false}
+WA_ENABLED=${5:-false}
+LIKE_PROD=${6:-$ENV}
+SHUTTERED=${7:-false}
 
 RUN_DIR=$(pwd)
 
 if [ -z "$TYPE" ] || [ -z "$VERSION" ] || [ -z "$ENV" ]; then
-    echo "Usage: create-xlsx.sh [type] [version] [env] [wa_enabled] [like_prod] [shuttered]"
+    echo "Usage: create-xlsx.sh [type] [version] [env] [use_identity] [wa_enabled] [like_prod] [shuttered] [use_identity]"
```
gemmatalbot: you've added [use_identity] twice, does the one at the end need to be there ?

KatKovacs1 (approved):

bin/create-xlsx.sh
nitinprabhuhmcts: good spot fixed

README.md
```diff
@@ -199,7 +199,7 @@ curl http://localhost:8008/health
 ### CCD config generation
 You can generate a CCD configuration with `create-xlsx.sh`.
 
-Usage: `create-xlsx.sh [type] [version] [env] [wa_enabled] [like_prod] [shuttered]`. The parameters `type`, `version` and
```
gemmatalbot, line 202: Do the example commands in the readme need to be updated?

gemmatalbot: Lines 208 to 226 I think

nitinprabhuhmcts: those are for local env and this flag is transparent as you don't need to specify it when running locally

gemmatalbot: You generate the file locally, but they are for generating ccd excel file for demo, aat and preview to upload 

e.g. 
Generate prod like CCD config for demo with WA turned off
./bin/create-xlsx.sh benefit dev demo false prod

nitinprabhuhmcts: updated readme now @gemmatalbot

gemmatalbot (approved):

PR #5440, opened 2026-09-01
Guard the NINO render in both halves of the bilingual PDF templates

### LINKED ISSUE #5438: NINO printed unguarded in one half of each bilingual PDF template (child-support appeals)
An appellant's National Insurance number is meant to be hidden on child-support PDFs. In the two
bilingual (Welsh/English) templates it is hidden in **one half of the document and printed in the
other** — and it's a different half in each file. A third template never hides it at all.

The English-only template gets it right, so the NINO appears for appellants who chose Welsh and not
for those who chose English.

## Where

`src/main/resources/templates/`

| template | prints NINO unguarded at | correctly guarded at |
|---|---|---|
| `evidenceDescriptionWelsh.html` | **line 111** (Welsh half) | line 192 (guard 186–195) |
| `personalStatementWelsh.html` | **line 182** (English half) | line 113 (guard 107–116) |
| `personalStatement.html` — English only | — | line 109 (guard from 103) |
| `onlineHearingSummary.html` | **line 137** | *no guard in this file* |

Four of the eight templates in that directory have a `hideNino` guard at all.

## Who it affects

`StorePdfService.java:202-204` — the flag is set for child-support appeals only:

```java
boolean hideNino = caseDetails.getData().getBenefitType()
        .filter(benefit -> benefit.equals(Benefit.CHILD_SUPPORT))
        .isPresent();
```

Lines 206-212 pass that flag into **both** the Welsh and the English `PdfAppealDetails`, so the
service is doing the right thing — the bug is entirely in the templates.

The value is live case data: `StorePdfService.java:199` reads
`caseDetails.getData().getAppeal().getAppellant().getIdentity().getNino()`. The templates in
`src/main/resources/` are the only copies (there is no `src/test/resources/templates/`), and none of
them contains a hardcoded NINO.

## Why the two halves drifted

Welsh in these templates is duplicated markup rather than a translation lookup, so every guard has
to be added twice by hand:

- `OldPdfService.createPdf` passes an `i18n` object to every template, but `grep -c i18n` across all
  8 templates returns 0 for every one.
- `I18nBuilder.build` unconditionally loads `/json/en.json`.
- There is no `cy.json` in the repository.

Each bilingual file is two hand-maintained copies of the same markup. Adding the missing guards
fixes the current behaviour but does not stop a future edit re-introducing the same gap.

## Suggested fix

1. Extend the guard to **both** halves of both bilingual templates, and add one to
   `onlineHearingSummary.html`.
2. Add a test asserting that in every template **every** `appealDetails.nino` render sits inside a
   `hideNino` guard — a count assertion (renders == guarded renders) rather than a spot-check, since
   the defect is exactly one render escaping in an otherwise-correct file.
3. Longer term, a real translation lookup for these templates so the two languages cannot diverge.

## Questions we could not answer from the source

1. Which user journeys render these templates, and so whether a Welsh child-support PDF has actually
   been produced. The flag and the templates are verified; the call graph from each entry point is
   not.
2. Whether any affected documents have been generated and sent.
3. `onlineHearingSummary.html` has no Welsh counterpart at all, so its unguarded render may affect
   every language rather than Welsh only. It looks like it belongs to a continuous-online-hearing
   flow (`AnswerState` carries COH wire strings) — if that flow is retired, it may affect nothing.

Found during an automated review of HMCTS platform configuration and re-verified against the source
by hand. Happy to provide anything else useful.

duncancrawford:
Fixes #5438.

An appellant's National Insurance number is meant to be hidden on child-support PDFs. Three renders
were not guarded:

| template | line | which half |
|---|---|---|
| `evidenceDescriptionWelsh.html` | 111 | the Welsh half |
| `personalStatementWelsh.html` | 182 | the English half |
| `onlineHearingSummary.html` | 137 | the only render in the file |

In the two bilingual templates the guard was present in one half and missing in the other — **and it
was a different half in each file** — so the NINO appeared for appellants who chose Welsh and not for
those who chose English.

`StorePdfService` is doing the right thing: it sets `hideNino` for child-support appeals and passes it
into both the Welsh and the English `PdfAppealDetails`. The gap was entirely in the templates.

## The change

Six lines. Each guard wraps the whole NINO `<div>`, matching the two templates that already had one:

```
{% if pdfSummary.appealDetails.hideNino == false  %}
<div>
    <dt class="cya-question">Rhif Yswiriant Gwladol</dt>
    <dd class="cya-answer">{{ pdfSummary.appealDetails.nino }}</dd>
</div>
{% endif %}
```

## Why a test, and why this shape

The Welsh templates are duplicated markup rather than a translation lookup — an `i18n` object is
passed to every template but no template references it, and there is no `cy.json` — so every guard
has to be added twice by hand and can drift apart again. `PdfTemplateNinoGuardTest` walks the guard
structure of every template and asserts that no `appealDetails.nino` render sits outside a `hideNino`
guard.

Two deliberate choices:

- **Structure, not counting.** A renders-equals-guards count would pass if a guard sat around some
  unrelated block. This tracks the open `{% if %}` stack, so it only passes if each render is
  genuinely inside one.
- **It asserts it found something.** `templatesScanned` and `rendersChecked` must both be positive,
  otherwise renaming the directory or the field would make the test pass silently — the failure a
  guard test is least able to notice about itself.

## Verification

Run against the **unfixed** templates the test fails and names exactly the three locations above;
against the fixed templates it passes. I could not run it through Gradle locally — the build needs a
Java 21 toolchain and this machine has 22 and 26, with no toolchain download repository configured —
so it was compiled and executed standalone on the JUnit 5 platform. It has no dependencies beyond
JUnit and AssertJ and reads only files, so that run is faithful, but **CI is the real check** and I'd
ask a reviewer to confirm it there.

## Left alone deliberately

- The four templates with no NINO render at all (`appellant_appeal_template.html`,
  `appellant_appeal_welsh_template.html`, `sent_notification.html`, and the letter templates) — the
  test covers them, so if a render is ever added without a guard it will fail.
- A real translation lookup for the bilingual templates, which is the actual fix for the drift. Worth
  doing, too large for this PR, and noted in #5438.

## One thing a reviewer can answer better than I can

`onlineHearingSummary.html` has no Welsh counterpart, so its unguarded render may have affected every
language rather than Welsh only. It looks like it belongs to a continuous-online-hearing flow
(`AnswerState` carries COH wire strings) — if that flow is retired, this part may be moot. I could not
establish from the source whether any affected documents were actually generated.

### DIFF
--- a/src/main/resources/templates/evidenceDescriptionWelsh.html
+++ b/src/main/resources/templates/evidenceDescriptionWelsh.html
@@ -103,6 +103,7 @@
                             {{ pdfSummary.appealDetails.surname }}
                         </dd>
                     </div>
+                    {% if pdfSummary.appealDetails.hideNino == false  %}
                     <div>
                         <dt class="cya-question">
                             Rhif Yswiriant Gwladol
@@ -111,6 +112,7 @@
                             {{ pdfSummary.appealDetails.nino }}
                         </dd>
                     </div>
+                    {% endif %}
                     <div>
                         <dt class="cya-question">
                             Cyfeirnod yr apêl
--- a/src/main/resources/templates/onlineHearingSummary.html
+++ b/src/main/resources/templates/onlineHearingSummary.html
@@ -129,6 +129,7 @@
                             {{ pdfSummary.appealDetails.surname }}
                         </dd>
                     </div>
+                    {% if pdfSummary.appealDetails.hideNino == false  %}
                     <div>
                         <dt class="cya-question">
                             NINO
@@ -137,6 +138,7 @@
                             {{ pdfSummary.appealDetails.nino }}
                         </dd>
                     </div>
+                    {% endif %}
                     <!--<div>
                       <dt class="cya-question">
                        Benefit type
--- a/src/main/resources/templates/personalStatementWelsh.html
+++ b/src/main/resources/templates/personalStatementWelsh.html
@@ -174,6 +174,7 @@
                             {{ pdfSummary.appealDetails.surname }}
                         </dd>
                     </div>
+                    {% if pdfSummary.appealDetails.hideNino == false  %}
                     <div>
                         <dt class="cya-question">
                             NINO
@@ -182,6 +183,7 @@
                             {{ pdfSummary.appealDetails.nino }}
                         </dd>
                     </div>
+                    {% endif %}
                     <div>
                         <dt class="cya-question">
                             Appeal reference number
--- a/src/test/java/uk/gov/hmcts/reform/sscs/service/pdf/PdfTemplateNinoGuardTest.java
+++ b/src/test/java/uk/gov/hmcts/reform/sscs/service/pdf/PdfTemplateNinoGuardTest.java
@@ -1,0 +1,79 @@
+package uk.gov.hmcts.reform.sscs.service.pdf;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.ArrayDeque;
+import java.util.ArrayList;
+import java.util.Deque;
+import java.util.List;
+import java.util.stream.Stream;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Every NINO render in a PDF template must sit inside a hideNino guard.
+ *
+ * <p>The Welsh templates are hand-duplicated markup rather than a translation lookup, so each
+ * guard has to be added twice. Three renders had escaped: the Welsh half of
+ * evidenceDescriptionWelsh, the English half of personalStatementWelsh, and the only render in
+ * onlineHearingSummary. A spot-check would not have found them, because in each case a sibling
+ * render in the same file was correctly guarded.
+ *
+ * <p>This walks the guard structure rather than counting, so a guard around some unrelated block
+ * cannot satisfy it.
+ */
+class PdfTemplateNinoGuardTest {
+
+    private static final Path TEMPLATES = Path.of("src/main/resources/templates");
+    private static final String NINO_RENDER = "pdfSummary.appealDetails.nino";
+    private static final String GUARD_SUBJECT = "hideNino";
+
+    @Test
+    void everyNinoRenderSitsInsideAHideNinoGuard() throws IOException {
+        List<String> unguarded = new ArrayList<>();
+        int rendersChecked = 0;
+        int templatesScanned = 0;
+
+        try (Stream<Path> templates = Files.list(TEMPLATES)) {
+            for (Path template : templates.filter(p -> p.toString().endsWith(".html")).toList()) {
+                templatesScanned++;
+                List<String> lines = Files.readAllLines(template);
+                Deque<Boolean> openGuards = new ArrayDeque<>();
+
+                for (int i = 0; i < lines.size(); i++) {
+                    String line = lines.get(i);
+
+                    if (line.contains("{% if ")) {
+                        openGuards.push(line.contains(GUARD_SUBJECT));
+                    }
+
+                    if (line.contains(NINO_RENDER) && !line.contains("{% if ")) {
+                        rendersChecked++;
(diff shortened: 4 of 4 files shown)


duncancrawford:
The red check on this PR is not the build. In run 33514093376:

- `buildMaster` — **success**
- `buildBranch` — **success**
- `report` — **failure**

The `report` job failed on its final step, posting the CCD diff as a PR comment:

```
##[error]Resource not accessible by integration
  https://docs.github.com/rest/issues/comments#create-an-issue-comment
```

`.github/workflows/ccd-diff.yaml` is triggered `on: pull_request`, and for a pull request **from a
fork** GitHub issues a read-only `GITHUB_TOKEN`. Creating a comment needs `issues: write`, which a
fork PR's token cannot be granted — adding a `permissions:` block would not change it. So this job
fails for any fork PR regardless of its contents, and this is the only fork PR in the last 60 in this
repo, which is why it has not come up before.

This change touches nothing under `definitions/benefit/sheets/`, so the diff the job computed is
empty — it is only the comment that could not be posted.

Two ways to get a fully green run, both needing a decision from someone with write access here:

1. Push this branch inside `hmcts/sscs-tribunals-case-api` and open the PR from there — the token
   then has write access and `report` passes. I do not have push access, which is why this came from
   a fork.
2. Make the job tolerate fork PRs, by skipping the comment step when the head repo is not this repo:

   ```yaml
   - if: github.event.pull_request.head.repo.full_name == github.repository
   ```

   or by giving it `continue-on-error: true`. Happy to raise that as its own issue rather than widen
   this PR, if it is wanted.

One note on the test in this PR: I could not run it through Gradle locally, because the build
requires a Java 21 toolchain and no toolchain download repository is configured, so on a machine with
only newer JDKs `./gradlew test` fails during configuration. I compiled and ran the test standalone on
the JUnit 5 platform instead — it has no project dependencies and only reads files — and verified it
both ways: against the unfixed templates it fails naming `evidenceDescriptionWelsh.html:111`,
`personalStatementWelsh.html:182` and `onlineHearingSummary.html:137`, and against the fix it passes.
CI remains the real check.

duncancrawford:
@nitinprabhuhmcts

duncancrawford:
Correcting something I implied above: **the green `build` check does not mean the new test ran.**

`.github/workflows/ci.yml` runs `./gradlew integration`, which does not include the unit `test` task,
so `PdfTemplateNinoGuardTest` was not executed by it.

Comparing the checks this fork PR received against a recent internal one (#5426):

| check | internal PR | this fork PR |
|---|---|---|
| `continuous-integration/jenkins/pr-head` | ✅ | absent |
| `Tests / Static checks / Container build / Unit tests and Sonar scan` | ✅ | absent |
| `Tests / Functional Test - preview` | ✅ | absent |
| `Tests / Smoke Test - AKS preview` | ✅ | absent |
| `Integration Tests results` | ✅ | absent |
| `build` / `buildBranch` / `buildMaster` | ✅ | ✅ |
| `report` | ✅ | ❌ (fork token, see #5441) |

Nine checks against four. Jenkins does not build pull requests from forks, and Jenkins is where the
unit tests and the checkstyle/PMD static checks run. So on this PR **neither my test nor any lint of
my Java has been exercised**, and I should not have let `build: SUCCESS` stand as if it had been.

What this means practically:

- The template change is six lines and the diff is above, so it can be read directly.
- The test is verified only by my own standalone run: compiled and executed on the JUnit 5 platform,
  failing against the unfixed templates with exactly `evidenceDescriptionWelsh.html:111`,
  `personalStatementWelsh.html:182`, `onlineHearingSummary.html:137`, and passing against the fix.
  That is real evidence but it is not this repo's toolchain, and it has not seen checkstyle at all.
- **The reliable way to get this properly tested is to push the branch inside this repository** rather
  than take it from my fork. Then Jenkins runs, the unit tests run, and the static checks run. I do not
  have push access here, so if a maintainer would rather re-push it internally than review a fork PR,
  that is the better route and I have no objection to this PR being closed in favour of it.

I would rather flag this than have the green ticks read as more assurance than they carry.

